PR #24, opened 2019-06-11
Added support  to use option defaults when not submitted by users

shreys7:
Implemented a check in every codegen to use option's default value if it isn't submitted by user, thus getOptions works as source of truth for defaults


codegens/curl/lib/index.js
```diff
@@ -1,21 +1,27 @@
 var sanitize = require('./util').sanitize,
     form = require('./util').form,
-    _ = require('./lodash');
+    _ = require('./lodash'),
+    self;
 
-module.exports = {
+self = module.exports = {
```
umeshp7: Why `self`? 
Can we directly use `this`? this.getOptions(); 😛

codegens/csharp-restsharp/lib/restsharp.js
```diff
@@ -124,9 +125,14 @@ module.exports = {
 
             //  snippet to create request in csharp-restsharp
             snippet = '';
+        self.getOptions().forEach((option) => {
+            if (_.isUndefined(options[option.id])) {
```
umeshp7: now the option[option.id] can be defined but be of a wrong type. I know we did not check this before but it could be a problem in future. Initially we might have assumed that we can never get wrong type, since it is tightly coupled with the UI. But now since we are trying to develop this independently, we could face this issue.
for example: if followRedirect could be a string, in that case this could be a problem. 
We can discuss more on this.

codegens/java-unirest/lib/index.js
```diff
@@ -1,58 +1,5 @@
...
-            name: 'Include Boilerplate',
-            id: 'includeBoilerplate',
-            type: 'boolean',
-            default: false,
-            description: 'Boolean denoting whether to include class definition and import statements in snippet'
-        }
-    ];
-}
-
 module.exports = {
```
umeshp7: this file is basically just exporting those functions? 
We could directly export them from the place they are defined and use `this` keyword to access getOptions function.

abhijitkane:
@shreys7 Comments to be addressed before merging

codegens/csharp-restsharp/lib/restsharp.js
abhijitkane: @umeshp7 The only thing we should check for is that our code doesn't throw exceptions for wrong option types.
Approving PR for now

codegens/csharp-restsharp/lib/util.js
```diff
@@ -1,19 +1,76 @@
...
     }
     inputString = inputString.replace(/\\/g, '\\\\').replace(/"/g, '\\"');
     return trim ? inputString.trim() : inputString;
 }
 
+/**
+ * sanitize input options
+ *
+ * @param {Object} options - Options provided by the user
+ * @param {Object} expectedOptions - Expected options built from getOptions
```
abhijitkane: the term `expected` only makes sense in tests

let's call this `options, defaultOptions`

codegens/csharp-restsharp/test/unit/convert.test.js
```diff
@@ -247,4 +249,62 @@ describe('csharp restsharp function', function () {
...
+        });
+    });
+
+    describe('sanitizeOptions function', function () {
+        var expectedOptions = {},
+            testOptions = {},
+            sanitizedOptions;
+
+        getOptions().forEach((option) => {
+            expectedOptions[option.id] = {
```
abhijitkane: defaultOption here too

codegens/csharp-restsharp/test/unit/convert.test.js
```diff
@@ -247,4 +249,62 @@ describe('csharp restsharp function', function () {
...
+                type: option.type
+            };
+            if (option.type === 'enum') {
+                expectedOptions[option.id].availableOptions = option.availableOptions;
+            }
+        });
+
+        it('should remove option not supported by module', function () {
+            testOptions.randomName = 'random value';
+            sanitizedOptions = sanitizeOptions(testOptions, expectedOptions);
```
abhijitkane: `expected` only makes sense if you're asserting against it

codegens/csharp-restsharp/test/unit/convert.test.js
```diff
@@ -247,4 +249,62 @@ describe('csharp restsharp function', function () {
...
+        it('should use defaults when option type is valid but value is invalid', function () {
+            testOptions = {};
+            testOptions.indentCount = -1;
+            testOptions.indentType = 'spaceTab';
+            testOptions.requestTimeout = -3000;
+            sanitizedOptions = sanitizeOptions(testOptions, expectedOptions);
+            expect(sanitizedOptions.indentCount).to.equal(expectedOptions.indentCount.default);
+            expect(sanitizedOptions.indentType).to.equal(expectedOptions.indentType.default);
+            expect(sanitizedOptions.requestTimeout).to.equal(expectedOptions.requestTimeout.default);
+        });
```
abhijitkane: don't see a test case for when testOptions has correct values

codegens/csharp-restsharp/lib/util.js
```diff
@@ -1,19 +1,79 @@
...
+                    if (typeof options[id] !== 'boolean') {
+                        result[id] = defaultOptions[id].default;
+                    }
+                    else {
+                        result[id] = options[id];
+                    }
+                    break;
+                case 'integer':
+                    if (typeof options[id] !== 'number' || options[id] < 0) {
+                        result[id] = defaultOptions[id].default;
```
abhijitkane: @shreys7 This seems wrong. Unless we know what option it is, we can't assume >0.
Can we change the type for indentCount and requestTimeout to `positiveInteger`, and update references?

PR #29, opened 2019-06-18
Added codegen php-pecl-http

shreys7:


codegens/php-pecl-http/.editorconfig
```diff
@@ -0,0 +1,14 @@
+# editorconfig.org
+root = true
```
abhijitkane: can we move .editorconfig to the root repo?
Also remove it from other folders if it exists

codegens/php-pecl-http/.eslintrc
```diff
@@ -0,0 +1,360 @@
+{
+  "plugins": [
```
abhijitkane: same for .eslintrc

codegens/php-pecl-http/.gitignore
```diff
@@ -0,0 +1,66 @@
+.DS_Store
+# Logs
+logs
```
abhijitkane: this should only have stuff not covered by the root .gitignore

codegens/php-pecl-http/.jsdoc-config.json
```diff
@@ -0,0 +1,34 @@
+{
+  "tags": {
```
abhijitkane: where do we use this file?

codegens/php-pecl-http/.npmignore
```diff
@@ -0,0 +1,74 @@
+### NPM Specific: Disregard recursive project files
+### ===============================================
+/.editorconfig
```
abhijitkane: same comment as .gitignore

PR #32, opened 2019-06-26
Updating options copy/names to be consistent across languages

abhijitkane:


test/codegen/structure.test.js
```diff
@@ -193,7 +193,8 @@ describe('Code-gen repository ' + codegen, function () {
     it('must have consistent type, description and name', function () {
       options.forEach((option) => {
         if (expectedOptions[option.id]) {
-          expect(option.description).to.be.eql(expectedOptions[option.id].description);
+          // the description property can be different across different languages
```
umeshp7: Even for same options? For e.g. `indentType` will have the same meaning across codegens and hence should have the same description. 
If it has a different meaning/description it should be a new option.

abhijitkane: indentType is the same, but request timeout is different across languages (seconds vs ms, for example).
Unless we're mandating that certain options must be present in all codegens, they should be free to define their own.

PR #36, opened 2019-07-02
Fixed requestTimeout issues 

shreys7:
Codegens affected : 

- Added Implementation (was present in getOptions, not actually being used) : 
  - restsharp
  - js-xhr
  - ruby
- Added support for indefinite timeout : 
  - golang (also converted from seconds to milliseconds)
  - java-unirest
  - shell-httpie
  - shell-wget
  - swift
- Added as a whole new option : 
  - libcurl


codegens/csharp-restsharp/lib/restsharp.js
```diff
@@ -17,7 +17,9 @@ function makeSnippet (request, options) {
     var snippet = `var client = new RestClient("${sanitize(request.url.toString())}");\n`,
         isUnSupportedMethod = UNSUPPORTED_METHODS_LIKE_GET.includes(request.method) ||
             UNSUPPORTED_METHODS_LIKE_POST.includes(request.method);
-
+    if (options.requestTimeout) {
```
abhijitkane: @shreys7 Can you confirm if not adding a timeout means the code waits indefinitely?

codegens/csharp-restsharp/test/unit/convert.test.js
```diff
@@ -179,7 +178,6 @@ describe('csharp restsharp function', function () {
                 options = {
                     indentCount: 1,
                     indentType: 'tab',
-                    requestTimeout: 2000,
```
abhijitkane: why have we removed this?

codegens/golang/lib/index.js
```diff
@@ -168,7 +168,10 @@ module.exports = {
         }
 
         if (timeout > 0) {
-            codeSnippet += `${indent}timeout := time.Duration(${timeout} * time.Second)\n`;
+            codeSnippet += `${indent}timeout := time.Duration(${timeout / 1000} * time.Second)\n`;
+        }
+        else {
+            codeSnippet += `${indent}timeout := time.Duration(0 * time.Second)\n`;
```
abhijitkane: Does 0 mean indefinite?
What happens if we don't add this line at all?

codegens/golang/test/unit/convert.test.js
```diff
@@ -125,7 +125,7 @@ describe('Golang convert function', function () {
                     options = {
                         indentCount: 1,
                         indentType: 'tab',
-                        requestTimeout: 2000,
+                        requestTimeout: 5000,
```
abhijitkane: why has this changed?

codegens/java-unirest/lib/unirest.js
```diff
@@ -6,19 +6,22 @@ var _ = require('./lodash'),
...
 function makeSnippet (request, indentString, options) {
     var snippet = '',
         urlString = encodeURI(request.url.toString());
 
     if (options.requestTimeout > 0) {
-        snippet += `Unirest.setTimeouts(${options.requestTimeout}, 0);\n`;
+        snippet += `Unirest.setTimeouts(0, ${options.requestTimeout});\n`;
+    }
+    else {
+        snippet += 'Unirest.setTimeouts(0, 0);\n';
```
abhijitkane: isn't the first arg supposed to be the read timeout?

codegens/shell-httpie/test/unit/converter.test.js
```diff
@@ -144,6 +144,31 @@ describe('Shell-Httpie convert function', function () {
       }
     });
   });
+
+  it.only('should add a timeout of 1 hour (3600 seconds) for RequestTimeout set to 0', function () {
```
abhijitkane: .only?

codegens/shell-httpie/test/unit/converter.test.js
```diff
@@ -144,6 +144,31 @@ describe('Shell-Httpie convert function', function () {
...
+      }),
+      options = {
+        requestTimeout: 0
+      };
+    convert(request, options, function (error, snippet) {
+      if (error) {
+        expect.fail(null, null, error);
+      }
+      expect(snippet).to.be.a('string');
+      expect(snippet).to.include('--timeout 3600');
```
abhijitkane: what happens if we don't add `timeout 3600`?

codegens/csharp-restsharp/lib/restsharp.js
shreys7: @abhijitkane The default timeout is 100 seconds. Adding the timeout as -1 is considered as infinite timeout. Will make the requrired changes.

codegens/csharp-restsharp/test/unit/convert.test.js
shreys7: @abhijitkane Previously it wasn't used in the code at all. After handling the timeout, some requests are getting bailed out. Will put it back by increasing the timeout to 5 seconds.

codegens/java-unirest/lib/unirest.js
shreys7:
```
Unirest.setTimeouts(long connectionTimeout, long socketTimeout);
Connection timeout: The timeout until a connection with the server is established (in milliseconds), and the default value is 10000
Socket timeout: The timeout to receive data (in milliseconds), and the default is 60000.
We can set any of these timeouts to zero to disable the timeout.
```

codegens/shell-httpie/test/unit/converter.test.js
shreys7: Will change this.

codegens/shell-httpie/test/unit/converter.test.js
shreys7: default is 30 seconds. We cannot provide an infinite value, hence 3600 seconds = 1 hour is set.

codegens/golang/test/unit/convert.test.js
shreys7: Previously it wasn't used in the code properly as it accepts input in seconds and was used directly . After handling the timeout, some requests are getting bailed out. Hence changed to 5 seconds.

codegens/golang/lib/index.js
shreys7: A Timeout of zero means no timeout.
The default is 0. They haven't provided anywhere in the official docs but read it on a github issue specifying to use different defaults. Will remove the else block.

Issue #50, opened 2019-07-12
Inconsistent semicolon usage in nodejs/request snippet


abhijitkane:
Seen in 0.0.9

```
var request = require('request');
var fs = require('fs')
var options = {
  'method': 'GET',
  'url': 'httpbin.org/get',
  'headers': {
  }
}
request(options, function (error, response) { 
  if (error) throw new Error(error);
  console.log(response.body);
});
```

PR #62, opened 2019-08-07
Fixed incorrect usage of semicolons in nodejs-request fixes

shreys7:
This pull request fixes issue #50


abhijitkane:
Test failing

codegens/nodejs-request/test/unit/snippet.test.js
```diff
@@ -263,6 +263,46 @@ describe('nodejs-request convert function', function () {
...
+      request = new sdk.Request(mainCollection.item[0].request);
+      options = {};
+      convert(request, options, function (error, snippet) {
+        if (error) {
+          expect.fail(null, null, error);
+        }
+        console.log(snippet);
+        expect(snippet).to.be.a('string');
+        var snippetArray = snippet.split('\n');
+        snippetArray.forEach(function (line, index) {
```
abhijitkane: hm...tests seem a little roundabout. Can we do an exact string comparison check instead, and comments on the part of the string we're actually testing for?

PR #64, opened 2019-08-07
Fixed removal of extra comma/s when requestTimeout is set in nodejs request

### LINKED ISSUE #63: Extra commas are added when requestTimeout option is set to non zero value in nodejs request
Seen in 0.0.9

```
var request = require('request');
var fs = require('fs');
var options = {
  'method': 'GET',
  'url': 'https://postman-echo.com/headers',
  'headers': {
    'my-sample-header': 'Lorem ipsum dolor sit amet',
    'not-disabled-header': 'ENABLED'
  },
  timeout: 1000,,
  followRedirect: false
};
request(options, function (error, response) { 
  if (error) throw new Error(error);
  console.log(response.body);
});
```

```
var request = require('request');
var fs = require('fs');
var options = {
  'method': 'GET',
  'url': 'https://postman-echo.com/headers',
  'headers': {
    'my-sample-header': 'Lorem ipsum dolor sit amet',
    'not-disabled-header': 'ENABLED'
  },
  timeout: 1000,
};
request(options, function (error, response) { 
  if (error) throw new Error(error);
  console.log(response.body);
});
```

shreys7:
This pull request fixes issue #63

### DIFF
--- a/codegens/nodejs-request/lib/request.js
+++ b/codegens/nodejs-request/lib/request.js
@@ -39,7 +39,7 @@ function makeSnippet (request, indentString, options) {
     );
   }
   if (options.requestTimeout) {
-    optionsArray.push(indentString + `timeout: ${options.requestTimeout},`);
+    optionsArray.push(indentString + `timeout: ${options.requestTimeout}`);
   }
   if (options.followRedirect === false) {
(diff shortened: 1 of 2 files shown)


abhijitkane:
@shreys7 Tests failing

PR #77, opened 2019-08-21
Updating copy for option name/descriptions for various codegens

abhijitkane:


codegens/http/lib/code-http-converter.js
```diff
@@ -9,17 +9,17 @@ let utils = require('./util');
  */
 function getOptions () {
   return [{
-    name: 'Body trim',
+    name: 'Trim request body fields',
     id: 'trimRequestBody',
     type: 'boolean',
     default: true,
-    description: 'Trim request body fields'
+    description: 'Remove white space and additional lines that may affect the server\'s response'
```
shreys7: Shall we have the same description for http too ? Stating 'that may affect the server\'s response'  doesn't really apply for http messages @abhijitkane

PR #80, opened 2019-08-23
Addressed unresolved variable encoding issues

shreys7:
Some of the codegens doesn't encode url on its own and fails to run if not encoded.
For such codegens, we have decided to encode everything in the url other than the unresolved variables
Languages affected: 
- curl
- swift
- java-unirest

### DIFF
--- a/codegens/curl/lib/index.js
+++ b/codegens/curl/lib/index.js
@@ -4,6 +4,40 @@ var sanitize = require('./util').sanitize,
   _ = require('./lodash'),
   self;
 
+/**
+ *
+ * @param {*} urlObject The request sdk request.url object
+ * @returns {String} final url string converted from parsing url object
+ */
+function getUrlStringfromUrlObject (urlObject) {
+  var url = '';
+  if (urlObject.protocol) {
+    url += (urlObject.protocol.endsWith('://') ? urlObject.protocol : urlObject.protocol + '://');
+  }
+  if (urlObject.auth && urlObject.auth.user) {
+    url = url + ((urlObject.auth.password) ?
+      // ==> username:password@
+      urlObject.auth.user + ':' + urlObject.auth.password : urlObject.auth.user) + '@';
+  }
+  if (urlObject.host) {
+    url += urlObject.getHost();
+  }
+  if (urlObject.port) {
+    url += ':' + urlObject.port.toString();
+  }
+  if (urlObject.path) {
+    url += urlObject.getPath();
+  }
+  if (urlObject.query && urlObject.query.count()) {
+    let queryString = urlObject.getQueryString({ ignoreDisabled: true, encode: true });
+    queryString && (url += '?' + queryString);
+  }
+  if (urlObject.hash) {
+    url += '#' + urlObject.hash;
+  }
+
+  return url;
+}
 self = module.exports = {
   convert: function (request, options, callback) {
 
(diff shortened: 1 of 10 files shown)


abhijitkane:
@shreys7 ruby is in this list too, right?

codegens/curl/lib/index.js
```diff
@@ -4,6 +4,40 @@ var sanitize = require('./util').sanitize,
   _ = require('./lodash'),
   self;
 
+/**
+ *
+ * @param {*} urlObject The request sdk request.url object
+ * @returns {String} final url string converted from parsing url object
+ */
+function getUrlStringfromUrlObject (urlObject) {
```
abhijitkane: @shreys7 This needs extensive unit tests

codegens/curl/lib/index.js
```diff
@@ -4,6 +4,40 @@ var sanitize = require('./util').sanitize,
...
+  var url = '';
+  if (urlObject.protocol) {
+    url += (urlObject.protocol.endsWith('://') ? urlObject.protocol : urlObject.protocol + '://');
+  }
+  if (urlObject.auth && urlObject.auth.user) {
+    url = url + ((urlObject.auth.password) ?
```
abhijitkane: @shreys7 Is this code copied from the postman-collection module? Any reason we can't use what's already there?

codegens/curl/lib/index.js
```diff
@@ -4,6 +4,40 @@ var sanitize = require('./util').sanitize,
...
+      urlObject.auth.user + ':' + urlObject.auth.password : urlObject.auth.user) + '@';
+  }
+  if (urlObject.host) {
+    url += urlObject.getHost();
+  }
+  if (urlObject.port) {
+    url += ':' + urlObject.port.toString();
+  }
+  if (urlObject.path) {
+    url += urlObject.getPath();
```
abhijitkane: does this always include the leading slash?

codegens/curl/test/unit/convert.test.js
```diff
@@ -252,5 +252,72 @@ describe('curl convert function', function () {
...
+          ]
+        }
+      });
+      options = {};
+      convert(request, options, function (error, snippet) {
+        if (error) {
+          expect.fail(null, null, error);
+        }
+        expect(snippet).to.be.a('string');
+        expect(snippet).to.include('http://postman-echo.com/post?a={{xyz}}');
```
abhijitkane: why can't we do an equality check here?

codegens/curl/lib/index.js
shreys7: but this is the implementation that is taken from underyling postman-collection sdk. Do we need to add tests for that too ? @abhijitkane

codegens/curl/lib/index.js
shreys7: Yes, this is the code copied from postman-collection module.
As we discussed, previously we were using `request.url.toString()` method which does all of this but **without encoding queryParams**. For this we stopped using that `.toString()` and copied the implementation of the same with just one tweak of setting encoding as true in `getQueryString({ encode: true })` method.

codegens/curl/lib/index.js
shreys7: this is what it is present in the underlying postman-collection implementation.

codegens/curl/test/unit/convert.test.js
shreys7: equality for the whole generated snippet ?

shreys7:
> @shreys7 ruby is in this list too, right?

No. I was basically trying to make a request as `https://postman-echo.com/{{method}}` and this was throwing error. But making a request as `https://postman-echo.com/post?a={{xyz}}` doesn't throw an error

PR #81, opened 2019-08-28
Fixed an issue where xhr.send(data) was being sent, even when request…

umeshp7:
….body was not present.

### DIFF
--- a/codegens/js-xhr/test/unit/convert.test.js
+++ b/codegens/js-xhr/test/unit/convert.test.js
@@ -242,6 +242,47 @@ describe('js-xhr convert function', function () {
       });
     });
   });
+
+  describe('Request with no body', function () {
+    var req = {
+        'method': 'GET',
+        'url': {
+          'raw': 'https://postman-echo.com/get',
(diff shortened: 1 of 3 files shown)


codegens/js-xhr/test/unit/convert.test.js
```diff
@@ -242,6 +242,47 @@ describe('js-xhr convert function', function () {
       });
     });
   });
+
+  describe('Request with no body', function () {
+    var req = {
+        'method': 'GET',
+        'url': {
+          'raw': 'https://postman-echo.com/get',
```
shreys7: Although it doesn't change the output, the path should be https://postman-echo.com/post @umeshp7

umeshp7: A GET call to https://postman-echo.com/post ?

umeshp7: Ah, missed the path. 
I changed the path property to 'get'. Since I am making a GET call to `https://postman-echo.com/get`.

PR #84, opened 2019-08-30
Removed changing of { to [ everywhere inside formdata JSON body to convert it to swift dictionary

shreys7:
Details: 
For sending the form data parameters, we used to form a swift dictionary from the request body JSON that we get from request object.
the code used was: `sanitize(JSON.stringify(body, null, 4), mode, trim)`
and sanitize for form data was: `inputString.replace(/{/g, '[').replace(/}/g, ']')`
Sanitize had this code because the swift dictionary used `[` and `]`.

Two problems:
1. Hardcoding paramater as 4 in `JSON.stringify()` doesn't take into account the indentation option we provide.
2. If parameter key or value contains a `[` or `]` , it was being replaced by `{` and `}` respectively.
Fixed the issue by manually parsing the request body to form the swift dictionary accordingly

### DIFF
--- a/codegens/swift/lib/swift.js
+++ b/codegens/swift/lib/swift.js
@@ -51,8 +51,26 @@ function parseURLEncodedBody (body, mode, trim) {
  * @returns {String} request body in the desired format
  */
 function parseFormData (body, mode, trim, indent) {
-  var parameters = sanitize(JSON.stringify(body, null, 4), mode, trim),
-    bodySnippet = `let parameters = ${parameters} as [[String : Any]]\n\n`;
+  var parameters = [],
+    parameter,
+    bodySnippet;
+  _.forEach(body, (data) => {
+    if (!(data.disabled)) {
+      parameter = '';
+      parameter += `${indent}[\n${indent.repeat(2)}"key": "${sanitize(data.key, mode, trim)}",\n`;
+      if (data.type === 'file') {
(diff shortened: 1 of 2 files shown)


abhijitkane:
@shreys7 Unit tests for the two functions?

shreys7:
> @shreys7 Unit tests for the two functions?

Basically this issue was encountered after the failing newman test for the formdata field key or value containing `{` or `}`. Hence no extra tests required @abhijitkane

codegens/swift/lib/swift.js
```diff
@@ -51,8 +51,26 @@ function parseURLEncodedBody (body, mode, trim) {
...
 function parseFormData (body, mode, trim, indent) {
-  var parameters = sanitize(JSON.stringify(body, null, 4), mode, trim),
-    bodySnippet = `let parameters = ${parameters} as [[String : Any]]\n\n`;
+  var parameters = [],
+    parameter,
+    bodySnippet;
+  _.forEach(body, (data) => {
+    if (!(data.disabled)) {
+      parameter = '';
+      parameter += `${indent}[\n${indent.repeat(2)}"key": "${sanitize(data.key, mode, trim)}",\n`;
```
abhijitkane: can be combined with the previous line

shreys7: we need to initialise the parameter variable every time with an empty string. Hence these two statements cannot be combined.

PR #87, opened 2019-09-03
Common newman tests

umeshp7:
Moved out newman tests to the root. 
Added util functions for newman tests for all codegens to consume.


codegens/curl/test/unit/convert.test.js
```diff
@@ -1,150 +1,54 @@
...
           }
-          runSnippet(snippet, collection, done);
+
+          return next(null, snippets);
         });
-      });
-    });
+      },
+      function (snippets, next) {
+        snippets.forEach((item, index) => {
```
abhijitkane: @umeshp7 If this code is common, can we add proper comments to this code in atleast one language, and point to the commented version in all the other languages?
Also, any way to re-use this too?

umeshp7: Yes, I will see if we can re-use this too. Meanwhile I will add a comment in curl.

umeshp7: Updated this. 
Now you just need to call ```runNewmanTest(convert, options, testConfig);```

PR #88, opened 2019-09-03
Removed unused require('fs') for nodejs-request

shreys7:
The line `var fs = require(''fs);` was being added in the snippet even when fs was not required at all. fs is only used to upload a file via form-data

### DIFF
--- a/codegens/nodejs-request/lib/request.js
+++ b/codegens/nodejs-request/lib/request.js
@@ -14,9 +14,18 @@ var _ = require('./lodash'),
  */
 function makeSnippet (request, indentString, options) {
   var snippet = 'var request = require(\'request\');\n',
-    optionsArray = [];
-
-  snippet += 'var fs = require(\'fs\');\n';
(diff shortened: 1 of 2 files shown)


codegens/nodejs-request/lib/request.js
```diff
@@ -14,9 +14,18 @@ var _ = require('./lodash'),
  */
 function makeSnippet (request, indentString, options) {
   var snippet = 'var request = require(\'request\');\n',
-    optionsArray = [];
-
-  snippet += 'var fs = require(\'fs\');\n';
+    optionsArray = [],
+    isFormDataFile = false;
+  if (request.body && request.body.mode === 'formdata') {
```
abhijitkane: why do we need .toJSON when reading `formdata` and not while reading `mode`?

shreys7: .mode is available before .toJSON but for reading the formdata array, we convert it to .toJSON for easy access

PR #92, opened 2019-09-04
Replaced usage of double quotes with single quotes in entire curl snippet

### LINKED ISSUE #72: For cURL escaping of body content, single quotes should be used to improve readability
The Postman app, without these code generators enabled, will escape the JSON body for a cURL request using single-quotes, which makes the body easily readable when copy/pasting the cURL example into docs or scripts. For example, here is what the normal Postman app might shoe for a cURL request that includes some JSON:
```json
curl -X POST \
  https://example.com/api/command \
  -H 'Accept: application/json' \
  -H 'Authorization: Basic TzJbGNtdA==' \
  -H 'Content-Type: application/json' \
  -H 'X-Request-ID: 1b3d12ad-e2f5-442e-99bb-bbea58936683' \
  -d '{
    "action": "query",
    "target": {
        "features": [
            "versions",
            "profiles",
            "pairs"
        ]
    }
}
'
```
Note that it is using single-quotes to pass the Headers and Body content to the cURL command-line. As such, there is very little backslash escaping and readability is high. Furthermore, by not adding all the double-quote escapes to the JSON body content, I can copy/paste it directly into something else, since it is just JSON (and not JSON with a bunch of weird escaping, which is not valid JSON).

If I enable this experimental code generator, the result is decidedly less readable, since the code-generators cURL support chooses to use double-quotes, causing much more unnecessary escaping of the JSON:
```json
curl -X POST "https://example.com/api/command" \
    -H "Authorization: Basic TzJbGNtdA==" \
    -H "Content-Type: application/json" \
    -H "Accept: application/json" \
    -H "X-Request-ID: 1b3d12ad-e2f5-442e-99bb-bbea58936683" \
    -d "{
    \"action\": \"query\",
    \"target\": {
        \"features\": [
            \"versions\",
            \"profiles\",
            \"pairs\"
        ]
    }
}"
```
If I copy/paste this JSON body content into jsonlint.com, you will see that it is no longer valid JSON due to all the extra backslash escaping.

As such, I'd like to revert the code-generators back to using the single-quote for sanitizing the headers and body content. I have a fix for this and will submit a pull request in the next couple weeks for review.

shreys7:
This fixes #72 
This will improve the readability of curl command as previously entire JSON body used to have double quotes and we used to escape each and every double quote as --data used double quotes to define data.

### DIFF
--- a/codegens/curl/lib/index.js
+++ b/codegens/curl/lib/index.js
@@ -37,15 +37,15 @@ self = module.exports = {
     }
     url = getUrlStringfromUrlObject(request.url);
     if (request.method === 'HEAD') {
-      snippet += ` ${form('-I', format)} "${url}"`;
+      snippet += ` ${form('-I', format)} '${url}'`;
     }
     else {
-      snippet += ` ${form('-X', format)} ${request.method} "${url}"`;
+      snippet += ` ${form('-X', format)} ${request.method} '${url}'`;
     }
 
     headersData = request.getHeaders({ enabled: true });
     _.forEach(headersData, function (value, key) {
-      snippet += indent + `${form('-H', format)} "${sanitize(key, trim)}: ${sanitize(value, trim)}"`;
+      snippet += indent + `${form('-H', format)} '${sanitize(key, trim)}: ${sanitize(value, trim)}'`;
     });
 
     if (request.body) {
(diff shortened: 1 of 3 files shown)


umeshp7:
@shreys7
Some conflicts to be resolved.

abhijitkane:
@shreys7 What's the update on this?

shreys7:
@abhijitkane Looked up various resources and everywhere it is mentioned that to escape single quotes inside single quotes, curl has this specification of using `'\''`

PR #95, opened 2019-09-08
Added long format options for different data sending methods in curl

shreys7:
Now using:
- --data-raw: for sending raw data( no special interpretation of `@` character)
- --data-urlencoded: for sending urlencoded body
- --data-binary: for sending binary data

### DIFF
--- a/codegens/curl/lib/index.js
+++ b/codegens/curl/lib/index.js
@@ -13,7 +13,7 @@ self = module.exports = {
     }
     options = sanitizeOptions(options, self.getOptions());
 
-    var indent, trim, headersData, body, text, redirect, timeout, multiLine,
+    var indent, trim, headersData, body, redirect, timeout, multiLine,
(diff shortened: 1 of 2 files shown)


abhijitkane:
@shreys7 Need a more detailed explanation / reason for this PR

shreys7:
@abhijitkane we were using --data or -d (short form) to send raw data as well as urlencoded data. Problems with using this:
1. @ has a special interpretation when used with --data, i.e. it expects a file path after it. Whereas if the user has selected raw body within postman, he clearly doesn't want this @ to be interpreted as a file upload but merely a character. curl specifies to use --data-raw if we don't want to have special interpretation of @ character.
2. Using --data for urlencoded has other cons. We were manually encoding characters and then sending this data using --data. Whereas curl specifies using --data-urlencode format for sending url encoded values, and it does encoding on its own.

PR #96, opened 2019-09-09
Added logic to trim header keys and not trim header values

shreys7:
Postman-App’s behavior:

- It doesn’t allow sending request which have whitespaces before or after header keys, i.e. throws an error in postman console stating invalid token in header key. Hence to stay in aligned with the app, we will be trimming header keys

- Postman allows sending whitespaces before or after header values. Confirmed after looking at the request in postman console as well as wireshark. Hence won’t be manipulating anything with header values, and remove everywhere where trimRequestBody option was used to trim header values also.


abhijitkane:
@shreys7 If this is also for the newman tests, the PR destination can be shifted

PR #104, opened 2019-09-18
Added the formdata fix in csharp-restsharp, and added dotnet compiler support

### LINKED ISSUE #100: Form-data POST request doesn't work as expected in csharp-restsharp
The below is the snippet generated for form data request:
```
var client = new RestClient("https://postman-echo.com/post");
client.Timeout = -1;
var request = new RestRequest(Method.POST);
request.AddParameter("a", "b");
request.AddParameter("c", "d");
IRestResponse response = client.Execute(request);
Console.WriteLine(response.Content);
```

Copy pasting this snippet and executing it actually sends the data as application/x-www-form-urlencoded instead of multipart/form-data.

shreys7:
This PR fixes #100


codegens/csharp-restsharp/lib/restsharp.js
```diff
@@ -30,6 +30,19 @@ function makeSnippet (request, options) {
...
+  if (request.body && request.body.mode === 'formdata') {
+    let isFile = false;
+    request.body.toJSON().formdata.forEach((data) => {
+      if (!data.disabled && data.type === 'file') {
+        isFile = true;
+      }
+    });
+    // The following statement needs to be added else the multipart/form-data request where there is no file
+    // is being sent as x-www-form-urlencoded by default
+    if (!isFile) {
```
umeshp7: Is this necessary to do? 
Will it be okay if we use  AlwaysMultipartFormData for all kind of formdata requests?

shreys7: whenever there is a file, this parameter isn't necessary. It converts to multipart/form-data itself. This parameter only needs to be added when you have only text fields.

PR #111, opened 2019-09-27
Added support for multiple headers with same key but with different values

shreys7:


codegens/curl/lib/index.js
```diff
@@ -51,10 +51,13 @@ self = module.exports = {
         value: 'text/plain'
       });
     }
-    headersData = request.getHeaders({ enabled: true });
-    _.forEach(headersData, function (value, key) {
-      snippet += indent + `${form('-H', format)} '${sanitize(key, true)}: ${sanitize(value)}'`;
-    });
+    headersData = request.toJSON().header;
+    if (headersData) {
+      headersData = _.reject(headersData, 'disabled');
```
abhijitkane: @shreys7 Is the property in the headers `enabled` or `disabled`? We were checking `{enabled:true}` earlier

abhijitkane: Why is this change needed, though?

abhijitkane: Is the SDK concatenating headers in `getHeaders`?

shreys7: 1. If we call `request.getHeaders()` here we provided the object `{enabled: true}`. Now, what this getHeaders() does is adds the headers with the same name in an array. Processing each and every value, i.e first checking if it is an array or not, and if it is, run another loop for adding the same keys multiple times with different values. Resulted in messy code.'
2. When we use request.toJSON().header, it doesn't do the array logic. It simply returns all the headers in key-value pairs. Thus easy for us code-wise to add in the snippets.
3. Yes, if we use request.toJSON().header, we get an array of objects, where disabled is set as true whenever a header is disabled. Hence, the changes.

PR #120, opened 2019-11-07
Added handling of graphql body mode in all codegens

shreys7:
In the request body, the graphql object is comprised of two strings, namely `query` and `variables.`. We will manually stringify it as one JSON object and then send it as raw body.
```JSON.stringify({ query: query, variables: variables})```

Need to discuss how this can be tested. Do we write unit tests in all the codegens stating whether the correct part of code is added in the generated snippet, or add a newman test in the basic collection? If we decide to go with the newman test, what graphql server will be used for the sample request?

### CODE REVIEW

File: codegens/csharp-restsharp/lib/parseRequest.js [hunk 1]
```diff
@@ -41,6 +41,24 @@ function parseContentType (request) {
...
+ */
+function parseGraphQL (requestBody) {
+  let query = requestBody.graphql.query,
+    graphqlVariables;
+  try {
+    graphqlVariables = JSON.parse(requestBody.graphql.variables);
+  }
+  catch (e) {
+    graphqlVariables = {};
+  }
```
umeshp7: Why is this required?

shreys7: To catch the error from JSON parsing. Just sending an empty object for variables. Not to send any undefined value or something.

codegens/csharp-restsharp/lib/restsharp.js
```diff
@@ -29,6 +29,12 @@ function makeSnippet (request, options) {
     snippet += 'client.FollowRedirects = false;\n';
   }
   snippet += `var request = new RestRequest(${isUnSupportedMethod ? '' : ('Method.' + request.method)});\n`;
+  if (request.body && request.body.mode === 'graphql' && !request.headers.has('Content-Type')) {
```
abhijitkane: is request.headers case-insensitive?

shreys7: yes. content-type or Content-Type, both the cases it returns the same

test/codegen/newman/fixtures/basicCollection.json
```diff
@@ -280,6 +280,32 @@
 			},
 			"response": []
 		},
+		{
```
umeshp7: We should also add a unit test for the codegens that do not run newman tests.
@shreys7

shreys7: on it

shreys7: added unit tests @umeshp7

PR #122, opened 2019-11-11
Fixed failing code generation in http for requests without body

shreys7:
- Create a GET request with no body. Save it to a collection and click view in web. It fails to generate snippets for HTTP.
- It fails for all the requests where no body is present.
- The following function tries to access request.body.mode without having a check for the body field to be present:
```function getHeaders (request) {
  let contentTypeIndex = _.findIndex(request.headers.members, { key: 'Content-Type' }),
    formDataHeader = `multipart/form-data; boundary=${FORM_DATA_BOUNDARY}`,
    headers = '';
 
  if (contentTypeIndex >= 0) {
    if (request.headers.members[contentTypeIndex].value === 'multipart/form-data' || request.body.mode === 'formdata') {
      request.headers.members[contentTypeIndex].value = formDataHeader;
    }
  }
 
  _.forEach(request.headers.members, (header) => {
    header.key = header.key.trim();
  });
  headers = convertPropertyListToString(request.headers, '\n', false);
  if (request.body.mode === 'formdata' && contentTypeIndex < 0) {
    headers += `Content-Type: ${formDataHeader}`;
  }
  return headers;
}
```
- Added checks for request.body in the above function and corresponding tests.

### DIFF
--- a/codegens/http/lib/util.js
+++ b/codegens/http/lib/util.js
@@ -169,7 +169,7 @@ function getHeaders (request) {
 
   if (contentTypeIndex >= 0) {
     if (request.headers.members[contentTypeIndex].value === 'multipart/form-data' ||
-      (request.bidy && request.body.mode === 'formdata')) {
+      (request.body && request.body.mode === 'formdata')) {
       request.headers.members[contentTypeIndex].value = formDataHeader;
     }
   }
--- a/codegens/http/lib/util.js
+++ b/codegens/http/lib/util.js
@@ -168,7 +168,8 @@ function getHeaders (request) {
     headers = '';
 
   if (contentTypeIndex >= 0) {
-    if (request.headers.members[contentTypeIndex].value === 'multipart/form-data' || request.body.mode === 'formdata') {
+    if (request.headers.members[contentTypeIndex].value === 'multipart/form-data' ||
+      (request.bidy && request.body.mode === 'formdata')) {
       request.headers.members[contentTypeIndex].value = formDataHeader;
     }
   }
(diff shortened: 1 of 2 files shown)


codegens/http/lib/util.js
```diff
@@ -168,7 +168,8 @@ function getHeaders (request) {
     headers = '';
 
   if (contentTypeIndex >= 0) {
-    if (request.headers.members[contentTypeIndex].value === 'multipart/form-data' || request.body.mode === 'formdata') {
+    if (request.headers.members[contentTypeIndex].value === 'multipart/form-data' ||
+      (request.bidy && request.body.mode === 'formdata')) {
```
umeshp7: What is bidy?

umeshp7: How is the test passing with this?

shreys7: Sorry. Corrected the typo. and the test is for the snippet generation should not fail if the request does not have a body. This is happening, the snippet is getting generated.

codegens/http/test/unit/converter.test.js
```diff
@@ -52,6 +52,29 @@ describe('Converter test', function () {
       expect(snippet).to.include('key_containing_whitespaces:   value_containing_whitespaces  ');
     });
   });
+  it.only('should generate snippets(not error out) for requests with no body', function () {
```
abhijitkane: .only!!

codegens/http/lib/util.js
```diff
@@ -168,7 +168,8 @@ function getHeaders (request) {
     headers = '';
 
   if (contentTypeIndex >= 0) {
```
umeshp7: The fix is inside this particular block which checks if the 'Content-Type' header is present.  
The unit test you have written does not pass this criterion of `contentTypeIndex >= 0`. Hence the fix is not being tested. That is why the initial typo wasn't detected.
@shreys7

shreys7: yes. Shall I change the tests?

umeshp7: Yes, you will have to add a 'Content-Type' header. Only then the unit test make sense.

PR #123, opened 2019-11-11
Added handling of multiple files in same form data field

shreys7:
- Issue:
Adding multiple files or zero files causes snippet generation to fail in the documenter.
This cannot be reproduced in the App.
- Reason:
The form data field in request body is an array of objects.
```
{
    key: 'param key'
    type: 'file' or 'text'
    value: Value of the param is type is text, else an empty string
    src: This property differs what we get in App versus what we get in documenter
}
```
- src property always is a string representing a file path in App. When there are multiple files in the same form data param, different params are created for each file path.
- While in documenter, this is an empty array for 0 files, a string for a single file and an array of strings for multiple files.
- Hence data.src.split does not work when the input is an array as opposed to string.
- Before calling the form data parsing function, make sure that it receives src param always a string. Thus, split the src array into different form data params with each containing different file paths.


umeshp7:
Need unit/newman test for this fix.

umeshp7:
@shreys7 Tests failing.

shreys7:
Added newman/unit tests @umeshp7

PR #128, opened 2019-11-19
No file fix for form-data 

shreys7:


codegens/csharp-restsharp/lib/restsharp.js
```diff
@@ -65,7 +65,7 @@ function makeSnippet (request, options) {
       }
     });
     formdata.remove((item) => {
-      return (item.type === 'file' && Array.isArray(item.src));
+      return (item.type === 'file' && (Array.isArray(item.src) || !item.src || typeof item.src !== 'string'));
```
umeshp7: This is too complex.

PR #130, opened 2019-11-22
Added the parseUrlVariable function to resolve url variables in HTTP

shreys7:
If there were variables in URL, they were not being resolved and left as it is.
e.g. /:action?a=1&b=2
Fixed this, now it resolves to /post?a=1&b=2


codegens/http/lib/util.js
```diff
@@ -117,6 +117,39 @@ function convertPropertyListToString (propertyList, joinUsing, includeDisabled =
...
+  }
+
+  variableArray.forEach(function (variableArrayElement) {
+    request.url.host.forEach(function (hostArrayElement, hostArrayElementIndex) {
+      if (hostArrayElement === ':' + variableArrayElement.key) {
+        request.url.host[hostArrayElementIndex] = variableArrayElement.value;
+      }
+    });
+
+    request.url.path.forEach(function (pathArrayElement, pathArrayElementIndex) {
```
abhijitkane: we only need to check for url.path, right? :path variables don't exist in the host or query params

shreys7: ah, yes. done

PR #150, opened 2020-01-09
Fixes #115

### LINKED ISSUE #115: Support ES6 syntax for Node { Unirest, Native, Request }
It is cumbersome to change syntax to make use of ES6 features in each request. 

Expose a code generator setting for the Node Framework that enables ES6 syntax

**Additional context**
For example: 

```
var request = require('request');
var options = {
  'method': 'GET',
  'url': '{{url}}/hourdata/:id?hour={{hour}}',
  'headers': {
  }
};
request(options, function (error, response) { 
  if (error) throw new Error(error);
  console.log(response.body);
});
```
becomes
```
const request = require('request');
let options = {
  'method': 'GET',
  'url': '{{url}}/hourdata/:id?hour={{hour}}',
  'headers': {
  }
};
request(options, (error, response) => { 
  if (error) throw new Error(error);
  console.log(response.body);
});
```

(This is just an example. I understand that there will be use cases that require a more un-opinionated syntax.)

jaymovaliya:
Added const in JS Snippets

### DIFF
--- a/codegens/js-fetch/lib/index.js
+++ b/codegens/js-fetch/lib/index.js
@@ -24,7 +24,7 @@ function redirectMode (redirect) {
  * @param {boolean} trim trim body option
  */
 function parseURLEncodedBody (body, trim) {
-  var bodySnippet = 'var urlencoded = new URLSearchParams();\n';
+  var bodySnippet = 'const urlencoded = new URLSearchParams();\n';
   _.forEach(body, function (data) {
     if (!data.disabled) {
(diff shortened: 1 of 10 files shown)


umeshp7:
var to const is not the only change that is required. Also, it would be better to have a fix for #115 for various languages separately. 
There is a PR which takes care of NodeJS Request, https://github.com/postmanlabs/postman-code-generators/pull/180. 
Closing this issue, for now, please feel free to create separate PRs if not already created.

Issue #153, opened 2020-01-21
Should `err` variable check or rename as `_`(underscore)


whthT:
**Describe the bug**
All defined variables must use in GoLang but `err` variable in this line not used.

**Expected code snippet and corresponding request**
We can add nil check on `err` variable.

**Screenshots**
![Adsız](https://user-images.githubusercontent.com/5617385/72806515-65c59600-3c66-11ea-8736-91f7d73b5a19.png)

Related Line: https://github.com/postmanlabs/postman-code-generators/blob/ad1587a9763e94b4c9a98ff7a9276f47cccd324f/codegens/golang/lib/index.js#L284

webholik:
Fixed by #276.

PR #162, opened 2020-02-22
Fix: Golang unused variable that causes errors (issue #153)

ttzztztz:
The `golang` force to use all defined variables, otherwise, the compiler will output an error and refuse to compile.

There're two ways to fix it.
The first is to `Printf` the error to the console, so that the user will know what had happened by reading the console stdout output. (I used this method)
The second is to ignore the unused variables by naming a variable to `_`, This method will ignore the error, and user will never know what happened, this may confuse some user when something wrong happened.

So, I just fix it by use the variable to output the error, and the snippet can pass golang's compiler.


umeshp7:
Need a unit test for this.

ttzztztz:
@umeshp7 A unit test was added, please review it, thank you!

webholik:
Closing as it is same as #276. Technically this one was filed first, but that one is ready to be merged.

Issue #158, opened 2020-02-19
User-Agent header set incorrectly in case of C#


peterbozso:
**Describe the bug**
Affected code-generator: csharp-restsharp
The problem is clearly described in this issue of RestSharp: https://github.com/restsharp/RestSharp/issues/787
In a nutshell, the code-generator currently sets the User-Agent like this:
```csharp
request.AddHeader("User-Agent", "Mozilla/5.0 (Windows NT 10.0; Win64; x64; rv:72.0) Gecko/20100101 Firefox/72.0");
```
But it's not effective at all, see issue referenced above. The solution is from the same issue as well. The correct way to set it is on the client level, like this:
```csharp
client.UserAgent = "Mozilla/5.0 (Windows NT 10.0; Win64; x64; rv:72.0) Gecko/20100101 Firefox/72.0"
```

**To Reproduce**
Generate C# code for a request where the User-Agent header is set, then execute the generated code. Observe that the User-Agent is set to `RestSharp x.x.x.x` instead of the one in the generated code.

webholik:
Fixed by #170.

PR #165, opened 2020-02-23
Fix issue #158

ttzztztz:
I fixed issue #158 by changing the C# RestSharp API .

As He says, the 
```c#
request.AddHeader("User-Agent", "Mozilla/5.0 (Windows NT 10.0; Win64; x64; rv:72.0) Gecko/20100101 Firefox/72.0");
```
doesn't work and will be overwritten to a default UA, the official solution is to use 
```C#
client.UserAgent = "Mozilla/5.0 (Windows NT 10.0; Win64; x64; rv:72.0) Gecko/20100101 Firefox/72.0";
```
Instead of Add Header.

So I just fix it by checking whether a Header's key equals to `User-Agent`, If so, output the `client.UserAgent = "..."`, otherwise, output the `request.AddHeader(..., ...)`.

So It will work and fix the issues.

### DIFF
--- a/codegens/csharp-restsharp/lib/parseRequest.js
+++ b/codegens/csharp-restsharp/lib/parseRequest.js
@@ -106,7 +106,11 @@ function parseHeader (requestJson) {
 
   return requestJson.header.reduce((headerSnippet, header) => {
     if (!header.disabled) {
-      headerSnippet += `request.AddHeader("${sanitize(header.key, true)}", "${sanitize(header.value)}");\n`;
+      if (sanitize(header.key, true).toLowerCase() === 'user-agent') {
+        headerSnippet += `client.UserAgent = "${sanitize(header.value)}";\n`;
+      } else {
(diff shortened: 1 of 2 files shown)


ankit-m:
@ttzztztz can you add a comment on the GH issue #158 saying that you are working on it.

ankit-m:
> @ttzztztz can you add a comment on the GH issue #158 saying that you are working on it.

Same for all issues you have created a PR for.

ttzztztz:
> @ttzztztz can you add a comment on the GH issue #158 saying that you are working on it.

OK, I added very detail descriptions on the background of the issues and my solution.
Thank you!

shreys7:
Hey @ttzztztz, thanks for the contribution. Can you add an appropriate unit test for this fix? 
You can refer the tests here:
https://github.com/postmanlabs/postman-code-generators/blob/develop/codegens/csharp-restsharp/test/unit/convert.test.js

PR #169, opened 2020-02-24
Add error management to go requests

zoek1:
Add support to error management and clean error output if the flag `fatal` is activated.

I decided manage the error instead of rename to `_` the `err` variable due to some new developer could have problems if some errors is raised. Also I added a flag called `fatal`, instead of go through panic  gives the user a clean error message (this is without the panic error that could be nosiy).

Reference: #153


umeshp7:
Need unit tests for this. 
https://travis-ci.com/postmanlabs/postman-code-generators/builds/151241964#L2633 
You need to add this new option in the list of expected options. For the above ^ test to pass. 
https://github.com/postmanlabs/postman-code-generators/blob/develop/test/codegen/structure.test.js

zoek1:
Thanks, i'll made the changes and ping you when is ready

zoek1:
Fixed the build, fix the lint errors was more hard than the implemented changes :sweat_smile: @umeshp7

webholik:
Closing, as #276 does this without adding an extra flag. We cannot add an extra option unilaterally as the support for it must be added in Postman app UI. We will take another look at it if there's more demand for this.

PR #171, opened 2020-02-24
Fix maven issues while downloading java dependencies

umeshp7:
central.maven.org is moved to https://repo1.maven.org. 
This is causing CI to fail. 
Fixes https://github.com/postmanlabs/postman-code-generators/issues/179


manjots1607:
@umeshp7 and @shreys7 I also worked on this issue and working to find root cause of timeout of "follow redirects" in ok-http for that I need to run the test locally but unable to do so.
I couldn't figure out why "npm test" command is not running. 

Can You List All the steps needed to take before running npm test command locally

nikhilmuz:
Not noticed this thus submitted duplicate PR #187 BTW my observations:
- 100 seconds works fine.
- I am not sure, whether including zip is better or just changing repo URL for maven dependencies is good enough. I think including zip will cause an increase in repo size unnecessarily.

nikhilmuz:
@manjots1607 it requires jar dependencies for okhttp to work. see ci requirements bash file for the list of dependencies required before running tests locally.

umeshp7:
@nikhilmuz 
Using repo1.maven.org is not a good idea since that could also change without notice (this name seems temporary too). 
It does increase the repo size but it's a one time expense. We could find a better solution in the future, with specific test environments for each code-gen for more rigorous tests. 

Also, increasing timeout time does not solve the primary issue. The time it is taking to finish execution could be a problem with the code we have generated. The aim is not to pass the test, but to produce accurate code-snippets.
I am trying a fix with mockbin's request, hope that works since I don't see any issues with the java-okhttp code-generated for redirects. 

@manjots1607 Please refer to the contributing guide, hit us up if need more clarification in that.

PR #175, opened 2020-02-24
try catch block syntax addition

darkCavalier11:
i added some try catch block for better error handling with throw error. pls review  @shreys7 @umeshp7


umeshp7:
The PR seems to have a lot of unnecessary changes which is not in the scope of this PR. Please create separate PRs for differents issues.

PR #176, opened 2020-02-26
Add: try catch block

### LINKED ISSUE #167: Exception handling for index functions
Add some level of exception handling(try-catch blocks) for functions inside `lib` folder.

VibhorCodecianGupta:
Resolves #167 
For description, check issue comments

### DIFF
--- a/lib/index.js
+++ b/lib/index.js
@@ -19,13 +19,20 @@ module.exports = {
 
     validCodegen.forEach((codegen) => {
       main = codegen.main;
-      if (typeof main.getOptions !== 'function') {
-        return callback('Codegen~getOptions: getOptions is not a function');
+      try {
+        if (typeof main.getOptions !== 'function') {
+          throw TypeError('Codegen~getOptions: getOptions is not a function');
+        }
+        if (!main.getOptions) {
+          throw ReferenceError('Codegen~convert: Could not find condegen corresponding to provided language, variant pair');
+        }
+      } catch(e) {
+        if (e instanceof TypeError || e instanceof ReferenceError) {
+          console.log(e.message)
+        } else {
+          console.log(e)
+        }
       }
-      if (!main.getOptions) {
-        return callback('Codegen~convert: Could not find condegen corresponding to provided language, variant pair');
-      }
-
       return callback(null, main.getOptions());
     });
   },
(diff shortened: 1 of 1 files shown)


VibhorCodecianGupta:
code snippet: https://www.joyent.com/node-js/production/design/errors
why it's bad practice:  https://www.joyent.com/node-js/production/design/errors#fn:1
Found this link useful to refer to @shreys7 @umeshp7

Added try catch block in index functions, the link above suggests using error handling methods like `throw/catch` with synchronous functions for capturing operational errors. It also explicitly mentions that invoking callbacks inside `catch` blocks is bad practice and ineffective which is why I did not replace the callback flow.

Does this issue need adding try/catch to index of separate codegens as well? Will do that if required

lib/index.js
```diff
@@ -19,13 +19,20 @@ module.exports = {
...
+      try {
+        if (typeof main.getOptions !== 'function') {
+          throw TypeError('Codegen~getOptions: getOptions is not a function');
+        }
+        if (!main.getOptions) {
+          throw ReferenceError('Codegen~convert: Could not find condegen corresponding to provided language, variant pair');
+        }
+      } catch(e) {
+        if (e instanceof TypeError || e instanceof ReferenceError) {
+          console.log(e.message)
```
umeshp7: Adding console.log for an error doesn't make sense here, if the `return` will be called anyway with `callback(null, main.getOptions());` 

The whole point of the check is to not call `getOptions()` if it is not a function.

lib/index.js
```diff
@@ -86,26 +93,35 @@ module.exports = {
...
+        throw ReferenceError('Codegen~convert: Could not find condegen corresponding to provided language, variant pair');
       }
-    });
-    if (!convert) {
-      return callback('Codegen~convert: Could not find condegen corresponding to provided language, variant pair');
-    }
 
+    } catch(e) {
+      if(e instanceof TypeError || e instanceof ReferenceError) {
+        console.log(e.message);
```
umeshp7: Same comment. 
Just by logging the error, the main issue is not solved. 
You need to return from inside the catch block itself.

lib/index.js
```diff
@@ -86,26 +93,35 @@ module.exports = {
   convert (language, variant, request, options, callback) {
     var convert, main;
 
-    if (!sdk.Request.isRequest(request)) {
-      return callback('Codegen~convert: Invalid request');
-    }
+    try {
+      if (!sdk.Request.isRequest(request)) {
+        throw TypeError('Codegen~convert: Invalid request');
```
umeshp7: If the provided request is not of the right sdk.Request type. Then there is no point in throwing error, you can simply callback with the error.

umeshp7:
Closing PR. 
@VibhorCodecianGupta Your reasoning is correct, try-catch should be used for operational errors but some of the checks made by the `if` blocks are essentials, For example: if the request is not of correct SDK request type, we don't need to call the convert function, since using a non-complaint request will cause convert function to fail. 
You need to identify parts of the code in the index file (if any), where try-catch could be added.

PR #180, opened 2020-02-27
Added ES6 feature support to NodeJS-Request codegen

sastava007:
**What it does**
It allows the user to generate the code snippet for NodeJS Request in ES6 format.
This is in response to issue #115 requested by one of the users.

**Changes I made**

- Registered a new parameter ES6_enabled in **`options()`** which allows the user to toggle between whether codegen generates snippet with ES6 features or not.

- By default ES6_enabled is set to `false`

- Also added the new created option parameter in **`structure.test.js`** file to register it as a valid option id.

- Added unit tests to verify the generated snippets.

**Follow Up**
This PR is first in the series of upcoming PRs which will add the feature of generating code snippet in ES6 notation with NodeJS.

### DIFF
--- a/codegens/nodejs-request/lib/request.js
+++ b/codegens/nodejs-request/lib/request.js
@@ -17,11 +17,12 @@ function makeSnippet (request, indentString, options) {
     optionsArray = [],
     isFormDataFile = false;
   if (options.ES6_enabled) {
-    snippet = 'let request = require(\'request\');\n';
+    snippet = 'const ';
   }
   else {
-    snippet = 'var request = require(\'request\');\n';
+    snippet = 'var ';
   }
+  snippet += 'request = require(\'request\');\n';
   if (request.body && request.body.mode === 'formdata') {
     _.forEach(request.body.toJSON().formdata, function (data) {
       if (!data.disabled && data.type === 'file') {
@@ -31,18 +32,20 @@ function makeSnippet (request, indentString, options) {
   }
   if (isFormDataFile) {
     if (options.ES6_enabled) {
-      snippet += 'let fs = require(\'fs\');\n';
+      snippet += 'const ';
     }
     else {
-      snippet += 'var fs = require(\'fs\');\n';
+      snippet += 'var ';
     }
+    snippet += 'fs = require(\'fs\');\n';
   }
   if (options.ES6_enabled) {
-    snippet += 'let options = {\n';
+    snippet += 'let ';
   }
   else {
-    snippet += 'var options = {\n';
+    snippet += 'var ';
   }
+  snippet += 'options = {\n';
 
   /**
      * creating string to represent options object using optionArray.join()
@@ -86,11 +89,12 @@ function makeSnippet (request, indentString, options) {
   snippet += optionsArray.join(',\n') + '\n';
   snippet += '};\n';
 
+  snippet += 'request(options, ';
   if (options.ES6_enabled) {
-    snippet += 'request(options, (error, response) => {\n';
+    snippet += '(error, response) => {\n';
   }
   else {
-    snippet += 'request(options, function (error, response) {\n';
+    snippet += 'function (error, response) {\n';
   }
   snippet += indentString + 'if (error) throw new Error(error);\n';
   snippet += indentString + 'console.log(response.body);\n';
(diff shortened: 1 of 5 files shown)


sastava007:
@shreys7 I have made all the changes as you said in your last review, hope it's fine now.

codegens/nodejs-request/lib/request.js
```diff
@@ -13,9 +13,15 @@ var _ = require('./lodash'),
  * @returns {String} - nodejs(request) code snippet for given request object
  */
 function makeSnippet (request, indentString, options) {
-  var snippet = 'var request = require(\'request\');\n',
+  var snippet,
     optionsArray = [],
     isFormDataFile = false;
+  if (options.ES6_enabled) {
```
shreys7: This is your code
```
if (options.ES6_enabled) {
    snippet = 'let request = require(\'request\');\n';
  }
  else {
    snippet = 'var request = require(\'request\');\n';
  }
```

It should be more like the one mentioned below(lesser code repetition):
```
if (options.ES6_enabled) {
    snippet = 'let ';
  }
  else {
    snippet = 'const ';
  }
snippet += 'request = require(\'request\');\n'
```

codegens/nodejs-request/lib/request.js
```diff
@@ -13,9 +13,15 @@ var _ = require('./lodash'),
...
+    snippet = 'let request = require(\'request\');\n';
```
shreys7: `const` instead of `let`

codegens/nodejs-request/lib/request.js
```diff
@@ -24,9 +30,19 @@ function makeSnippet (request, indentString, options) {
     });
   }
   if (isFormDataFile) {
-    snippet += 'var fs = require(\'fs\');\n';
+    if (options.ES6_enabled) {
```
shreys7: same comment as above. Combine for lesser code repetition

codegens/nodejs-request/lib/request.js
```diff
@@ -24,9 +30,19 @@ function makeSnippet (request, indentString, options) {
...
+      snippet += 'let fs = require(\'fs\');\n';
```
shreys7: `const` instead of `let`

sastava007: @shreys7 Last time, I used const so you asked me to use var/ let. That's why I changed.

shreys7: no. Those changes for if-else statements. We should use const for fs and request.

sastava007:
@shreys7 Have updated as you said, once this one gets accepted for **`request`** I'll fix the **`native`** and **`unirest`** with the same approach.

sastava007:
@umeshp7 Can you please review this?

codegens/nodejs-request/lib/request.js
```diff
@@ -120,6 +145,13 @@ function getOptions () {
       type: 'boolean',
       default: false,
       description: 'Remove white space and additional lines that may affect the server\'s response'
+    },
+    {
+      name: 'Generate the code snippet with ES6 features',
+      id: 'ES6_enabled',
+      type: 'boolean',
+      default: false,
+      description: 'Allows the user to generate code snippet with latest EcmaScript 6 (ES6) features'
```
umeshp7: This description is user-facing. So we don't need to specify 'user'. We have to come up with something that the user will read. 
We could go with something like: "Modifies code snippet to incorporate ES6 (EcmaScript) features."

codegens/nodejs-request/lib/request.js
```diff
@@ -120,6 +145,13 @@ function getOptions () {
       type: 'boolean',
       default: false,
       description: 'Remove white space and additional lines that may affect the server\'s response'
+    },
+    {
+      name: 'Generate the code snippet with ES6 features',
```
umeshp7: This could be simply 'Enable ES6 features'.

umeshp7:
We'll need an extra set of newman tests, with this option enabled true. We need to make sure the snippet generated is runnable.

sastava007:
@umeshp7 I have added a new block inside the main mocha suite with **`ES6_enabled`** : `true`, please have a look.

shreys7:
@sastava007 Merging this, you can start working on `native` and `unirest`. Thanks for the contribution.

Issue #172, opened 2020-02-24
Fixture is not able to find the file path in case of formdataFileCollection


sastava007:
**Describe the bug**
The **`formdataFileCollection`** fixture is not able to locate the exact path where `Newman` is trying to upload the files when testing the `codegens` locally.

**To Reproduce**
Run **`npm test nodejs-request`**

**Expected code snippet and corresponding request**
This can be solved by making the test.sh script to first create the required files with sample data and then running `addPathToFormdataFile.js` file.

**Screenshots**
![image](https://user-images.githubusercontent.com/43990819/75152615-766a9f80-572f-11ea-8425-bf93c39374d5.png)

**Additional context**
I would like to work on this issue.

PR #173, opened 2020-02-24
Move test file creation step from ci-requirements script to test script for form data testing

shreys7:
Previously, the files required for testing form data file uploads were created using `ci-requirements.sh` step. This file is not run locally and hence would cause form-data tests to fail locally. This step has been removed from `ci-requirements.sh` file and moved to `test.sh` file.
This fixes #172

PR #181, opened 2020-02-27
Fixed test file path for formdataFileCollection

sastava007:
**What it does**
This PR fixes the issue #172 , which was causing the **`formdataFileCollection`** test to fail as the test file that was required to upload by the codegen does not exist.

**Changes I made**

- Added command to generate test files in **`test.sh`** with sample data before beginning the test, and once test is completed they must be deleted. 

- Removed it's equivalent command from **`ci-requirements.sh`** which was causing redundancy.

**Follow Up**
`N/A`

@umeshp7  @shreys7  Please review!

### DIFF
--- a/npm/ci-requirements.sh
+++ b/npm/ci-requirements.sh
@@ -1,12 +1,5 @@
 #!/bin/bash
 set -ev; # stop on error
-echo "Creating test files and adding paths to collection for testing form data file uploads"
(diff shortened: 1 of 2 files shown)


umeshp7:
Duplicate of https://github.com/postmanlabs/postman-code-generators/pull/173

PR #187, opened 2020-02-29
Fixed Issue causing Travis Build fail

nikhilmuz:
Changes
- Updated maven repo URLs in CI requirements bash script for Travis.
- Mocha timeout implementation increased to 100 seconds for okhttp causing timeout error.

### DIFF
--- a/codegens/java-okhttp/test/newman/newman.test.js
+++ b/codegens/java-okhttp/test/newman/newman.test.js
@@ -2,6 +2,7 @@ var runNewmanTest = require('../../../../test/codegen/newman/newmanTestUtil').ru
   convert = require('../../lib/index').convert;
 
 describe('convert for different request types', function () {
+  this.timeout(10000); // Mocha timeout implementation
   var options = {indentCount: 3, indentType: 'Space', includeBoilerplate: true},
     testConfig = {
       compileScript: 'javac -cp *: main.java',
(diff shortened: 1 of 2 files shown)


nikhilmuz:
@umeshp7 Approval required. Compare it with #171 before merging.
Primary differences are:
- Time duration is 100 sec rather than 120 sec in mentioned PR.
- I have updated the broken URLs rather than including them in repository close this if that is more suitable.

umeshp7:
@nikhilmuz
Using repo1.maven.org is not a good idea since that could also change without notice (this name seems temporary too).
It does increase the repo size but it's a one time expense. We could find a better solution in the future, with specific test environments for each code-gen for more rigorous tests.

Also, increasing timeout time does not solve the primary issue. The time it is taking to finish execution could be a problem with the code we have generated. The aim is not to pass the test, but to produce accurate code-snippets.
I am trying a fix with mockbin's request, hope that works since I don't see any issues with the java-okhttp code-generated for redirects.

nikhilmuz:
@umeshp7 I was wondering why we ain't using some SDK by maven to get links?

umeshp7:
We would be interested in that. Any such SDK in your mind that you suggest? @nikhilmuz

nikhilmuz:
@umeshp7 [Apache Maven Download](http://maven.apache.org/download.cgi) is this what you are referring to? Actually it is included in the project for dependency management and an XML file keeps list of dependency and version. I should create a feature request in GitHub issues I guess if it is worth a try.

umeshp7:
Feel free to create a PR for this fix, we would be happy to review. 

Taking this discussion on the above issues and closing this thread.

Issue #174, opened 2020-02-24
Convert bash scripts(.sh files) to javascript files


shreys7:
Currently, the following files present inside `npm` folder are bash scripts

- ci-requirements.sh
- package.sh
- test.sh
- zip.sh

Change these bash scripts to javascript files, so that it becomes platform agnostic, i.e. works well on Windows, Linux or macOS. This can be achieved using the node's [shelljs](https://www.npmjs.com/package/shelljs) module.
See `deepinstall.js` file present in `npm` folder for reference.

sastava007:
@shreys7 
I'm working on this.

manjots1607:
@shreys7 
I also want to work on this issue.

apshada:
@sastava007  & @manjots1607  are you still working on this issue, if not then i would like to take this issue

sastava007:
Thanks for your interest, @apshada. 
I have almost completed it, and will push by tonight.

debck:
@shreys7 I would like to work on the issue.

PHAGUN-JAIN:
@shreys7 if this issue is resolved, it should be closed sir!

PR #198, opened 2020-03-06
ci-requirements.js file replaced with ci-requirements.sh file

manjots1607:
What this PR does
This PR converts ci-requirements.sh file to ci-requirements.js file. This is in response to issue #174.

Changes you made
I deleted the previously written bash file and written a .js file using shelljs module to execute bash commands.
updated "npm run cirequirements" command in package.json, so as .js file would run when it is called. 

Screenshot/ CodeSnippet
Not required


manjots1607:
@umeshp7 @shreys7
Please review

sastava007:
@manjots1607 Have not I mentioned that I'm already working on this?

umeshp7:
There are conflicts. Please rebase with the latest develop. 
Also there are indentation differences in these changes as compared to the rest of the repo. Please fix those after rebasing with the latest develop.

manjots1607:
Thanks For Review, I will resolve issues as soon as possible.

manjots1607:
Unexpected error in travis CI please check @umeshp7

manjots1607:
Fixed Error by replacing:-
"npm install shelljs" to "npm install"
in Travis CI.
please review it again @umeshp7 .

manjots1607:
Please review it @umeshp7 @shreys7

webholik:
I am not convinced that we need to replace the `ci-requirements.sh` with javascript. It will not make the code more portable - we are still running `apt-get` which is only available on Debian based Linux distributions. On the other hand it causes problem if you are running the script on your local machine. `shelljs` does not provide a `tty` to commands so there is no way to give interactive input to them. This can be problem, for example, for `java-okhttp` codegen, if you already have the required jar files installed, the `unzip` command will ask interactively if you want to replace them. There is no way to send a reply so the script just hangs there. 

I think we should stick to the shell script for now until we have a better solution.

webholik:
Closing this as ci-requirements.sh will be kept as bash script.

PR #204, opened 2020-03-11
Try catch for lib/index

### LINKED ISSUE #167: Exception handling for index functions

VibhorCodecianGupta:
fixes #167 

@umeshp7 @shreys7 another attempt.
Please review whether things work this time around. Also, I found a lot of files inside `lib` of individual codegens which could use try-catch too. Shall we do it?


lib/index.js
```diff
@@ -11,23 +11,28 @@ module.exports = {
...
-      var lang = codegen.lang.trim(),
-        currentVariant = codegen.variant.trim();
-      return language === lang.toLowerCase() && variant.toLowerCase() === currentVariant.toLowerCase();
-    });
+    try {
+      var validCodegen = languageMap.filter((codegen) => {
+        var lang = codegen.lang.trim(),
+          currentVariant = codegen.variant.trim();
+        return language === lang.toLowerCase() && variant.toLowerCase() === currentVariant.toLowerCase();
+      });
```
gamblegoon: lib/index.js

gamblegoon: thank you team for your point of directions

VibhorCodecianGupta: Is there a review here? I didn't get you

webholik (commented):
Passing errors in callback's first argument is standard pattern in NodeJs. This is also documented in the code-generators README. Throwing an exception should be reserved for situations where there is no scope for recovery. 

For `try catch` blocks we need to know which errors we are  expecting and how to recover when we get them. Just putting entire block of code in try and console logging the exception is not how it should be done. 

Since it's been a long time since this PR was created, I will be closing this one.

PR #219, opened 2020-03-17
Converted package.sh and zip.sh to Javascript files.

sastava007:
**What this PR does**
This PR converts the `package.sh`and `zip.sh` bash script file to javascript making it the platform agnostic. This PR is last in the series of PRs trying to solve issue #174 

**Changes you made**
Converted the bash commands into it's closely equivalent JS commands using `shellJS` npm package.

**Screenshot/ CodeSnippet**
Not required

**Test Configuration**:
Tried to test locally for various codegens by running `npm package <codegen-name>` and comparing the result obtained with that generated from the similar bash script.

**Follow up**
Not required

*Thanks for contributing!* ❤️


sastava007:
@umeshp7 Can you please review?

npm/package.js
```diff
@@ -0,0 +1,97 @@
...
+// ensure that the working tree is clean before packaging
+commandOut = shell.exec('source ./npm/package/require_clean_work_tree.sh');
+if (commandOut.code !== 0) {
+    shell.exit(1);
+}
+/* commandOut = shell.exec('require_clean_work_tree "create package"');
+if (commandOut.code !== 0) {
+    console.log('Working tree is not clean, please commit/stash your changes');
+    shell.exit(1);
+} */
```
umeshp7: You can remove this, no need to keep commented code.

npm/package.js
```diff
@@ -0,0 +1,97 @@
...
+        function (next) {
+            console.log(`Generating zip for codegen ${codegen}`);
+            commandOut = shell.exec(`npm run zip ${codegen} --color always`, {silent: true});
+            if(commandOut.code !==0) {
+                console.error(`Failed to zip codegen ${codegen}, here\'s the error:`);
+                return next(commandOut.stderr);
+            }
+            console.log(commandOut.stdout);
+        }
+    ], (err) => {
```
umeshp7: This can be simply changed to `npm run test` (As done in the package.sh script). Where npm test would handle the testing of all languages. This could will be duplicated in test.js as well. Which will be difficult to maintain in case we decide to do some changes in one of them.

npm/zip.js
```diff
@@ -0,0 +1,33 @@
...
+try {
+    exists=fs.statSync(codegen_path).isDirectory();
+} catch (err) {
+    console.log(`Codegen ${codegen} doesn't exist, please enter a valid name`);
+    shell.exit(1);
+}
+if (exists) {
+    console.log(`${codegen} : zip`);
+    shell.pushd(codegen_path, "q");
+    commandOut = shell.exec('npm pack --color always');
```
umeshp7: What is the need of `--color always`?

npm/package.js
sastava007: @umeshp7 Yeah, we can do that. But in that case we have to assume that the user has already run the deepinstall and now he just wants to test for all the codegens.
And even in `package.sh` we are first running deepinstall and then going for test.
image

npm/zip.js
sastava007: @umeshp7 Because by default shellJS doesn't display the result of exec() in colors as done by Chalk. 
This has been highlighted here https://github.com/shelljs/shelljs/issues/86

npm/package.js
umeshp7: That's okay. That's a reasonable assumption. 
You can run tests on a new repo without doing `npm i` anyway. Here the user is us/developers I think we can safely assume that deepinstall will be run before.

sastava007: Okay, I'll do that then.

sastava007:
@umeshp7 Have updated, please look.

npm/package.js
sastava007: Done

sastava007:
@umeshp7 Any updates?

umeshp7:
@sastava007 Hey, I ran the scripts locally, and firstly I could find many lint errors. Can you please rebase with the recent develop and fix those. (Recent develop will have linting for root files as well) 

Also, remove `{ silent: true }` from every where.

sastava007:
@umeshp7 Please approve.

sastava007:
@umeshp7 Have fixed everything as you said, please have a look.

PR #223, opened 2020-03-19
Fixed example code and removed typos throughout the repo

sastava007:
**What this PR does**
This PR fixes the spelling mistakes that have been unknowingly left while development.

**Changes you made**

- Modified few sample code which were provided as an example.
- Fixed typos throughout the repository.

**Screenshot/ CodeSnippet**
N/A

**Test Configuration**:
N/A

**Follow up**
N/A

*Thanks for contributing!* ❤️

PR #225, opened 2020-03-23
Update README.md

someshkoli:
Example snippet had improper declaration of convert function and request instance.
Fixed and tested


sastava007:
@someshkoli This has been fixed with #223

PR #234, opened 2020-03-30
README typo fix

ajwad-shaikh:
A small typo fix in README that I encountered while studying the repository.


ajwad-shaikh:
I added some missing supported languages to README

Issue #249, opened 2020-05-17
Cannot run Python-Requests tester because of JSON errors


Marti2203:
**Describe the bug**
Clearly mention the code-generators that are affected by this issue.
While trying to clone this repo and run the tester to see that all tests run successfully, me and my teammate (@dasyad00) saw how the crashed, because of a JSON error. I have traced it down to the fact that python binary strings have a b at the start when printed out.

**To Reproduce**
1. Create fresh clone of repo
2. run ```npm install``` in terminal and ```npm run deepinstall dev;``` 
3. run ```npm test python-requests```
4. see where tests fail (JSON parse error)
**Expected code snippet and corresponding request**
If you have a working code snippet that you think solves your problem, please add it here:
Also, add the request.json for which this code snippet was generated.

**Screenshots**
If applicable, add screenshots to help explain your problem. Please make sure the screenshot does not contain any sensitive information such as API keys or access tokens. 
For example you could add the screenshot of the snippet/request builder from Postman app.

**Additional context**
If you have any reference material which might help us debug the problem please add it here:
Version of postman-code-generators/Postman app:
Latest version of master branch in github.
The error resides in the [newmanTestUtil.js line 87](https://github.com/postmanlabs/postman-code-generators/blob/develop/test/codegen/newman/newmanTestUtil.js#L87)

dasyad00:
Please find below the log files.
[log.txt](https://github.com/postmanlabs/postman-code-generators/files/4640505/log.txt)

someshkoli:
@Marti2203  @dasyad00  I guess this is due the python version and the byte stream. The test fails for python 3.x and passes for 2.x. Anyway I'm working on this and it will be fixed soon.

PR #250, opened 2020-05-17
Fix Python request multipart/form-data

dasyad00:
Currently, if you want to send form-data without any files, it does not work properly with `python-requests`. After a discussion with the developer of `python-requests-toolbelt`, @Marti2203 and I believe that using the aforementioned library would make things easier.


dasyad00:
Build has failed due to issue #249

webholik:
For generated code, we would ideally want no dependency on an external library other than `requests`. Could you file a bug for the issue you are facing with `form-data`?

Issue #186, opened 2020-02-29
Curl code from postman doesn't work


forenheith:
I created request at postman, tested it, all works fine, then I clicked on Code button at the right top corner, copied CURL code, and tried to use it in Windows Shell and got the errors

`At line:2 char:3
+ --header 'x-pr-tenantid: 2720F16E-1E5B-4AA5-93CB-6169E36B5C9E' \
+   ~
Missing expression after unary operator '--'.
At line:2 char:3
+ --header 'x-pr-tenantid: 2720F16E-1E5B-4AA5-93CB-6169E36B5C9E' \
+   ~~~~~~
Unexpected token 'header' in expression or statement.
At line:3 char:3
+ --header 'x-pr-apikey: <<<apikey>>> ...
+   ~
Missing expression after unary operator '--'.
At line:3 char:3
+ --header 'x-pr-apikey: <<<apikey>>> ...
+   ~~~~~~
Unexpected token 'header' in expression or statement.
At line:4 char:3
+ --header 'Authorization: Bearer <<auth>> ...
+   ~
Missing expression after unary operator '--'.
At line:4 char:3
+ --header 'Authorization: Bearer <<auth>> ...
+   ~~~~~~
Unexpected token 'header' in expression or statement.
    + CategoryInfo          : ParserError: (:) [], ParentContainsErrorRecordException
    + FullyQualifiedErrorId : MissingExpressionAfterOperator`

Then I changed --header on -h and got the new ones

`Invoke-WebRequest : A positional parameter cannot be found that accepts argument '--request'.
At line:1 char:1
+ curl --location --request GET 'http://localhost:60513//sip/transactio ...
+ ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
    + CategoryInfo          : InvalidArgument: (:) [Invoke-WebRequest], ParameterBindingException
    + FullyQualifiedErrorId : PositionalParameterNotFound,Microsoft.PowerShell.Commands.InvokeWebRequestCommand

-h : The term '-h' is not recognized as the name of a cmdlet, function, script file, or operable program. Check the spelling of the name, or if a path was included, verify that the path is correct and try again.
At line:2 char:1
+ -h 'x-pr-tenantid: 2720F16E-1E5B-4AA5-93CB-6169E36B5C9E' \
+ ~~
[...]

PR #251, opened 2020-05-17
Adds additional option to curl codegen

### LINKED ISSUE #248: cURL snippets fail on Windows due to Single Quote
**Describe the bug**
@umeshp7 @shreys7 
#72 #92 
this causes a problem on Windows machines; by pasting the cURL generated code snipped into a Windows Command Prompt, you get the misleading error message:

> curl: (1) Protocol "'https" not supported or disabled in libcurl

Which is just due to the single quotes, and can be fixed only by using double quotes.

see here: https://stackoverflow.com/questions/6884669/curl-1-protocol-https-not-supported-or-disabled-in-libcurl

Much like you have the option to change the line continuation character from \ to ^ in the Settings there, I think the single vs. double quote should also be an option.

As a bonus, would be nice if Postman detected that the machine is Windows and defaulted to ^ and " in the settings accordingly.

**To Reproduce**
Paste any resulting cURL snippet into a win cmd prompt

someshkoli:
Fixes #248 
Alternative to #92 
Adds `quoteType` option to curl codegen. 
Since windows does not accept single quoted url parameter, this option allows to set it to double quote. 
While adding the url to the snippet instead of hardcoding the quote `'` I have used this option to set the quote as per requirements. I haven't added the option paramere for all occurrence of `quote`. Added just for the url parameter.


codegens/curl/README.md
```diff
@@ -34,7 +35,8 @@ var request = new sdk.Request('www.google.com'),  //using postman sdk to create
         trimRequestBody: true,
         multiLine: true,
         followRedirect: true,
-        longFormat: true
+        longFormat: true,
+        quoteType: '\''
```
abhijitkane, line 39: the quoteType option here should be `singlequote` or `doublequote`. This can be translated to `'` or `"` inside the module.

codegens/curl/lib/index.js
```diff
@@ -186,6 +186,15 @@ self = module.exports = {
         description: 'Set a character used to mark the continuation of a statement on the next line ' +
           '(generally, \\ for OSX/Linux, ^ for Windows)'
       },
+      {
+        name: 'Quote Type',
+        id: 'quoteType',
+        availableOptions: ['\'', '"'],
+        type: 'enum',
+        default: '\'',
+        description: 'Set a Quote type to be used(Single or Double) to be used,' +
```
abhijitkane, line 195: This copy needs to be looked at. I'd suggest using https://www.grammarly.com/ to proofread any copy. Also, does this need to be different from what's mentioned in the readme?

codegens/curl/test/unit/convert.test.js
```diff
@@ -33,6 +33,50 @@ describe('curl convert function', function () {
...
+      convert(request, options, function (error, snippet) {
+        if (error) {
+          expect.fail(null, null, error);
+        }
+
+        snippetArray = snippet.split(' ');
+        expect(snippetArray[4][0]).to.equal('\'');
+      });
+    });
+    it('should return snippet with url in double quote(") as line continuation ' +
```
abhijitkane, line 58: what does the quote have to do with the line-continuation?

codegens/curl/lib/index.js
```diff
@@ -186,6 +186,15 @@ self = module.exports = {
         description: 'Set a character used to mark the continuation of a statement on the next line ' +
           '(generally, \\ for OSX/Linux, ^ for Windows)'
       },
+      {
+        name: 'Quote Type',
+        id: 'quoteType',
+        availableOptions: ['\'', '"'],
```
umeshp7, line 192: availableOptions should be ['singleQuote', 'doubleQuote'] as mentioned by @abhijitkane above.

codegens/curl/test/unit/convert.test.js
someshkoli: Ah its a typo nothing much

codegens/curl/lib/index.js
someshkoli: Fixed

codegens/curl/lib/index.js
someshkoli: Fixed this

codegens/curl/README.md
someshkoli: Fixed

someshkoli:
@shreys7 @umeshp7  we have request for three variations of curl command, one we have option for setting line breaker, now this one ^ for quotes and in this #186 requesting changes for curl to run in powershell. Can we just one option as "OS" or "shell type" ?? that would be much easier operate rather than having 3 different options which is anyway not gonna be used in all permutations ?

umeshp7:
@webholik Has many conflicts

webholik:
Bungled it up. New PR at #398.

PR #256, opened 2020-05-26
updates readme to include nodejs axios

r3dsm0k3:
Added nodejs axios in the readme.


r3dsm0k3:
@shreys7 I think this should also go along with the axios merge.

PR #276, opened 2020-06-01
Adds snippet to check err before using res in GoNative

someshkoli:
Fixes: https://github.com/postmanlabs/postman-app-support/issues/8318 partially
The err due to absence of `err` check is caused by go vet(kind of linting for go), which forces strict formatting, this includes checking for err being nil or not  if we have assigned some value to `err`. 
So basically if anyone uses go vet then `err` checking is mandatory if it has been assigned:
`if err!=nil {
    //do something
 }
`

 - adds err check after client.Do(req)
 - adds println and return statement inside err condition body

### DIFF
--- a/codegens/golang/lib/index.js
+++ b/codegens/golang/lib/index.js
@@ -280,10 +280,13 @@ self = module.exports = {
       codeSnippet += `${indent}req.Header.Set("Content-Type", writer.FormDataContentType())\n`;
     }
     responseSnippet = `${indent}res, err := client.Do(req)\n`;
+    responseSnippet += `${indent}if err != nil {\n${indent.repeat(2)}fmt.Println(err)\n`;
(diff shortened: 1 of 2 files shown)


abhijitkane:
@someshkoli The original issue also mentions using http/https as a URL prefix. Also, can you update your comment to include the before/after snippet?

someshkoli:
> @someshkoli The original issue also mentions using http/https as a URL prefix. Also, can you update your comment to include the before/after snippet?

before:
image

after:
image


Also I was not sure about the http/https thing as that can also be done through the postmanapp so is a check required anyway ? Like we encounter error of invalid url in nodejs-request too when https/http is not provided in the url ?

webholik (approved):

umeshp7 (approved):

PR #277, opened 2020-06-02
Fixes issue with semicolon in form data (CURL)

### LINKED ISSUE #257: Curl request Fail when form-data have semicolond in it
**Describe the bug**
Curl snippets fail when form data have semicolon(;) in it. 

**To Reproduce**
  - Create any request with form-data containing semi-colon(;)

According to curl documentation the form values which have semicolon or comma (; or ,) should be enclosed within qoutes.
image

**Screenshots**
image

**Additional context**
enclosing the contents inside quotes is helpfull but it comes with a downside. The quotes become part of the form data and are sent together

someshkoli:
fixes #257 
Curl requires form data to be enclosed in double quotes if form data value contains semicolon or comma
image

This PR adds check for the above conditions and adds the following changes 

- adds check for semicolon and comma in form data
 - encloses form data value if semicolon/comma is present
 - Modifies sanitize function, adds follows
    - adds option to escape double quotes
    - adds option to escape back slash

### DIFF
--- a/codegens/curl/lib/index.js
+++ b/codegens/curl/lib/index.js
@@ -141,16 +141,11 @@ self = module.exports = {
               if (!(data.disabled)) {
                 if (data.type === 'file') {
                   snippet += indent + `${form('-F', format)}`;
-                  snippet += data.value.match(/[,;]/g) ?
-                    ` '${sanitize(data.key, trim)}=@"${sanitize(data.src, trim, true, true)}"'` :
-                    ` '${sanitize(data.key, trim)}=@${sanitize(data.src, trim)}'`;
+                  snippet += ` '${sanitize(data.key, trim)}=@"${sanitize(data.src, trim, true, true)}"'`;
                 }
                 else {
                   snippet += indent + `${form('-F', format)}`;
-                  snippet += data.value.match(/[,;]/g) ?
-                    // eslint-disable-next-line max-len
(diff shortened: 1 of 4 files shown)


umeshp7:
https://travis-ci.com/github/postmanlabs/postman-code-generators/builds/169343297#L5074 
Newman test for formdata is failing. 
Please make sure the changes you have done are correct and works in all cases.

umeshp7:
> Curl requires form data to be enclosed in double quotes if form data value contains semicolon or comma

Where are we checking if semicolon and/or comma are present in form data?

umeshp7:
According to the image you shared, the `, and ;` have to be wrapped around double quotes if it occurs in a filename. Not for all kinds formdata values.

someshkoli:
If we do that it conflicts with " inside the form data.
Also if we do that, form data values having `"test"`  result in false respose, it takes string which is only present inside the quotes and does not consider quotes as a part of string. And if we try to escape those we get a whole new response from the backend. I have tried and tested the above approach and all the cases works with this

someshkoli:
> Where are we checking if semicolon and/or comma are present in form data?
> That approach fails for strings containing " so had to enclose all the form data in double quotes

webholik:
Problem is that behavior of `curl` has changed in newer versions. Quoting for non-file data is not required in version `7.47.0` which is what CI runs and adding quotes breaks that build as then the output after running `curl` no longer matches the one in `newmanResponses.json`.

umeshp7 (approved):

Issue #284, opened 2020-06-14
Request URL containing $ doesn't work well with PHP - cURL


honeyyadav:
# Short Explanation

When I am generating **PHP - cURL** code from a postman request with URL containing a **$** sign followed by a **letter** or **underscore** (for example $firstName, $_address, $LastName), then the code generated by postman doesn't work. This is because PHP treats them as variable name (but they are not variable name(s) in this case).

# Example

My request URL is : -
```
> http://example.com/path?$filter=name eq John
```
The code generated by postman : -
```php
<?php

$curl = curl_init();

curl_setopt_array($curl, array(
  CURLOPT_URL => "http://example.com/path?$filter=name%20eq%20John",
  CURLOPT_RETURNTRANSFER => true,
  CURLOPT_ENCODING => "",
  CURLOPT_MAXREDIRS => 10,
  CURLOPT_TIMEOUT => 0,
  CURLOPT_FOLLOWLOCATION => true,
  CURLOPT_HTTP_VERSION => CURL_HTTP_VERSION_1_1,
  CURLOPT_CUSTOMREQUEST => "GET",
));

$response = curl_exec($curl);

curl_close($curl);
echo $response;

```

The above code will not work as PHP will treat **\$filter** as a variable name. To solve this problem we need to put a **\\** (backslash) before **\$filter** like this : -

```php
<?php

$curl = curl_init();

curl_setopt_array($curl, array(
  CURLOPT_URL => "http://example.com/path?\$filter=name%20eq%20John",
  CURLOPT_RETURNTRANSFER => true,
  CURLOPT_ENCODING => "",
  CURLOPT_MAXREDIRS => 10,
  CURLOPT_TIMEOUT => 0,
  CURLOPT_FOLLOWLOCATION => true,
  CURLOPT_HTTP_VERSION => CURL_HTTP_VERSION_1_1,
  CURLOPT_CUSTOMREQUEST => "GET",
));

$response = curl_exec($curl);

curl_close($curl);
echo $response;

```

And now it will work **flawlessly**.

> **Postman version used:** v7.26.0

erdemkose:
Instead of escaping `$` characters, replacing double quotes(`"`) with single quotes(`'`) would be a simpler and cleaner solution. PHP does not look for variables in single-quoted strings.

shellstrom:
[...]

PR #285, opened 2020-06-14
Adding '\' before a valid PHP variable name solves #284

honeyyadav:
## Solves #284

# Why my solution works :-

My solution is very simple and straight forward, I am just replacing the occurrence of a valid PHP URL with '**\\**' added at the begning. I have tested this code with various test cases and it is working flawlessly.


erdemkose:
Instead of escaping `$` characters, replacing double quotes(`"`) with single quotes(`'`) would be a simpler and cleaner solution. PHP does not look for variables in single-quoted strings.

webholik:
Fixed by #352.

PR #289, opened 2020-06-22
Add sanitising of quotes in URLs(and paths)

shreys7:
Fixes https://github.com/postmanlabs/postman-app-support/issues/8674
Single/double quotes were not handled correctly in some of the languages.

Languages affected:
 - cURL
 - Java Unirest
 - NodeJS Native (path)
 - Python http.client
 - Swift

### DIFF
--- a/codegens/curl/lib/util.js
+++ b/codegens/curl/lib/util.js
@@ -1,4 +1,4 @@
-module.exports = {
+var self = module.exports = {
   /**
      * sanitizes input string by handling escape characters eg: converts '''' to '\'\''
      * and trim input if required
@@ -149,7 +149,7 @@ module.exports = {
       url += '#' + urlObject.hash;
     }
 
-    return url;
+    return self.sanitize(url);
   },
 
   /**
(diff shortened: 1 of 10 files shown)


abhijitkane:
@shreys7 were the sanitize functions already written? Also, what would be the correct curl command for `curl http://httpbin.org/get?a=a\"b\'c`?

shreys7:
> @shreys7 were the sanitize functions already written? Also, what would be the correct curl command for `curl http://httpbin.org/get?a=a"b'c`?

So, we already had sanitize functions, which basically takes care of escaping a single quote inside a single quote, and the same for double quotes. The sanitization part was present in headers, bodies, etc. But some languages did not have sanitisation of URLs.
for `curl http://httpbin.orf/get?a=a"b'c` => we put the URL in quotes, and after proper escaping cURL command would look like `curl 'http://httpbin.org/get?a=a"b'\''c'`

**Note**: to escape a single quote within a single quote we need to replace a single quote with `'\''` for cURL(shell)

PR #320, opened 2020-08-16
Change nodejs-axios snippet to use async/await

### LINKED ISSUE #302: Convert JS ES6 code to use async/await functions
No problem, just updating code.

I would like to convert functions to use async/await functions.

Example existing code output for nodejs-axios:
```javascript
const axios = require('axios');
let data = JSON.stringify({"field":"value"});

let config = {
  method: 'post',
  url: 'https://example.com/api/v1/endpoint',
  headers: { 
    'Authorization': 'Bearer auth_token', 
    'Content-Type': 'application/json'
  },
  data : data
};

axios(config)
.then((response) => {
  console.log(JSON.stringify(response.data));
})
.catch((error) => {
  console.log(error);
});
```

Proposed output:
```javascript
const axios = require('axios');
let data = JSON.stringify({"field":"value"});

const config = {
  method: 'post',
  url: 'https://example.com/api/v1/endpoint',
   headers: {
   'Authorization': 'Bearer auth_token',
   'Content-Type': 'application/json'
  },
  data: data
};

async function makeCall(config){
  try {
    const response = await axios(config);
    console.log(JSON.stringify(response.data));
  } catch (error) {
    console.log(error);
  }
}

makeCall(config);
```

VallariAg:
resolves #302 
`nodejs-axios` snippet now include async/await.

Snippets for `javascript` don't use async/await as well but there is no `ES6_enabled` option, so I wasn't sure if I should change the snippets to use features of ES6 by default, so I didn't change them. 

(I would love to contribute something more significant to Postman)

### DIFF
--- a/codegens/nodejs-axios/lib/axios.js
+++ b/codegens/nodejs-axios/lib/axios.js
@@ -132,23 +132,26 @@ function makeSnippet (request, indentString, options) {
   snippet += ' config = {\n';
   snippet += configArray.join(',\n') + '\n';
   snippet += '};\n\n';
-  snippet += 'axios(config)\n';
   if (options.ES6_enabled) {
-    snippet += '.then((response) => {\n';
+    snippet += 'async function makeCall(config) {\n';
+    snippet += indentString + 'try {\n';
+    snippet += indentString + indentString + 'const response = await axios(config);\n';
+    snippet += indentString + indentString + 'console.log(JSON.stringify(response.data));\n';
+    snippet += indentString + '} catch (error) {\n';
+    snippet += indentString + indentString + 'console.log(error) {\n';
+    snippet += indentString + '}\n';
+    snippet += '}\n'
+    snippet += 'makeCall(config);'
   }
   else {
+    snippet += 'axios(config)\n';
     snippet += '.then(function (response) {\n';
-  }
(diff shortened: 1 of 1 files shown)


webholik:
async/await was introduced in ES7 so technically it shouldn't be used for the ES6_enabled. Plus I think a lot of people will prefer the version using promises (last place I worked at had the policy of only using promises). Perhaps we could come up with an option to use async/await, and use it for all javascript and nodejs variants.

max-schumacher-shippeo:
Using async/await for the axios code generator is a great idea/improvement - I believe this should move forward as this syntax is now a widely used standard and preferable to callbacks.

umeshp7:
@VallariAg Any updates on adding this as an option?

VallariAg:
Hey @umeshp7. Oh sorry - I completely missed the comments here! 
Adding an option for this sounds superb. At the moment I'm loaded with work, can I pick this up later? If this is of high priority, someone else can pick this one up :+1: 
Sorry for the delay! :')

umeshp7:
@VallariAg No worries. Let us know if you can take this up in the coming week. Otherwise we'll assign it internally. Thanks for the contribution so far.

VallariAg:
@umeshp7 It's better if you could assign it internally. I'll checkout other ways to contribute to postman later :)

PR #343, opened 2020-09-21
added nodejs typesript axios

### LINKED ISSUE #342: Add support for typescript node js axios
Add support for typescript node js axios

Add codegen for typescript node js axios

**Additional context**

Abhirup-99:
Fixes #342


Abhirup-99:
I need some help in passing the build. The test file would create a ts file but how do I compile it down to js to ensure it works fine.

Abhirup-99:
the pr is ready for review.

dhwaneetbhatt (changes requested):
@ Abhirup-99 It would be better to pass an option to generate TypeScript code in the existing NodeJS-Axios generator rather than write an entirely new generator. Could you please update your fork with the latest changes from upstream and make the changes?

PR #352, opened 2020-10-06
Replace double-quotes by single-quotes in codegen/php-curl

webholik:
Solves #284 

PHP expands variables and escape sequences for special characters inside double-quotes. This is a problem if body of a request contains `$` sign. Using single-quotes solves this problem. This is an alternative solution to #285 .


umeshp7:
Can you add a newman test/ unit test with `$` present in the body? @webholik

umeshp7 (approved):

PR #354, opened 2020-10-09
Add optional import of FoundationNetworking in swift codegen to make it work with Swift 5.x

### LINKED ISSUE #353: Swift codegen needs to import FoundationNetworking to work with Swift 5.x
Starting with Swift 5.1, networking related functionality has been moved to `FoundationNetworking`. It needs to be imported for the codegen to work.

**To Reproduce**
Compile the code generated for any request with Swift 5.x and the following error is produced:
``` 
error: 'URLSession' is unavailable: This type has moved to the FoundationNetworking module. Import that module to use it.
```

The following code should be present in the generated code to work with Swift 5.x:
```
#if canImport(FoundationNetworking)
import FoundationNetworking
#endif
```

webholik:
Fix #353.

### DIFF
--- a/codegens/swift/lib/swift.js
+++ b/codegens/swift/lib/swift.js
@@ -335,7 +335,8 @@ self = module.exports = {
     requestBody = (request.body ? request.body.toJSON() : {});
     bodySnippet = parseBody(requestBody, trim, indent);
 
-    codeSnippet = 'import Foundation\n\n';
+    codeSnippet = 'import Foundation\n';
+    codeSnippet += '#if canImport(FoundationNetworking)\nimport FoundationNetworking\n#endif\n\n';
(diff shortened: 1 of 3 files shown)


umeshp7:
Can we test this somehow using the newman test?
What is the current version of swift that we are using in newman test?
@webholik

webholik:
@umeshp7 We are running Swift 5.0.1 for newman tests. The changes happened in Swift 5.1.  Perhaps we could change `ci-requirements.sh` to run Swift 5.1.

umeshp7:
@webholik yes we can update it to 5.1 
Maybe in future we can decide how we can run newman test for multiple versions.

umeshp7 (approved):

PR #392, opened 2020-10-23
Add ` as line continuation delimiter for curl

webholik:
Fixes #186

PR #393, opened 2020-10-23
Implement option shellType for curl codegen

### LINKED ISSUE #186: Curl code from postman doesn't work

### LINKED ISSUE #248: cURL snippets fail on Windows due to Single Quote

webholik:
Fixes #248 and fixes #186. Alternative to #251 and #392. 

We currently have an option called 'lineContinuationCharacter" to choose between \ and ^ as line continuation character for Unix shells and Windows cmd.exe respectively. But powershell uses ` as for escaping newlines, this is the cause of #186. These shells also differ in how they treat single quote and double quotes. Specifically, cmd treats single quotes as an ordinary character and passes it as it is to curl, which is the cause of #248. 

Instead of adding separate option for these, I have added a shellType option so the user can simply choose which shell he is going to run curl in and we will generate the correct code for that.

### DIFF
--- a/codegens/curl/lib/index.js
+++ b/codegens/curl/lib/index.js
@@ -15,7 +15,7 @@ self = module.exports = {
     options = sanitizeOptions(options, self.getOptions());
 
     var indent, trim, headersData, body, redirect, timeout, multiLine,
-      format, snippet, silent, url;
+      format, snippet, silent, url, shellType, lineContinuationCharacter, quoteType;
 
     redirect = options.followRedirect;
     timeout = options.requestTimeout;
@@ -23,6 +23,22 @@ self = module.exports = {
     format = options.longFormat;
     trim = options.trimRequestBody;
     silent = options.silent;
+    shellType = options.shellType;
+
+    switch (shellType) {
+      case 'cmd.exe':
+        lineContinuationCharacter = '^';
+        quoteType = '"';
+        break;
+      case 'powershell':
+        lineContinuationCharacter = '`';
+        quoteType = '\'';
+        break;
+      default:
+        lineContinuationCharacter = '\\';
+        quoteType = '\'';
+        break;
+    }
 
     snippet = silent ? `curl ${form('-s', format)}` : 'curl';
     if (redirect) {
@@ -33,17 +49,17 @@ self = module.exports = {
     }
     if (multiLine) {
       indent = options.indentType === 'Tab' ? '\t' : ' ';
-      indent = ' ' + options.lineContinuationCharacter + '\n' + indent.repeat(options.indentCount); // eslint-disable-line max-len
+      indent = ' ' + lineContinuationCharacter + '\n' + indent.repeat(options.indentCount); // eslint-disable-line max-len
     }
     else {
       indent = ' ';
     }
     url = getUrlStringfromUrlObject(request.url);
     if (request.method === 'HEAD') {
-      snippet += ` ${form('-I', format)} '${url}'`;
+      snippet += ` ${form('-I', format)} ${quoteType + url + quoteType}`;
     }
     else {
-      snippet += ` ${form('-X', format)} ${request.method} '${url}'`;
+      snippet += ` ${form('-X', format)} ${request.method} ${quoteType + url + quoteType}`;
     }
 
     if (request.body && !request.headers.has('Content-Type')) {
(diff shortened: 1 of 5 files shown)


codegens/curl/lib/index.js
```diff
@@ -15,14 +15,30 @@ self = module.exports = {
...
-      format, snippet, silent, url;
+      format, snippet, silent, url, shellType, lineContinuationCharacter, quoteType;
 
     redirect = options.followRedirect;
     timeout = options.requestTimeout;
     multiLine = options.multiLine;
     format = options.longFormat;
     trim = options.trimRequestBody;
     silent = options.silent;
+    shellType = options.shellType;
```
umeshp7, line 26: We should still check if `lineContinuationCharacter` is present in the option object and choose the shell type accounting that in too. And if the `lineContinuationCharacter` is chosen as `^` we should use shell type `cmd.exe` This might cause problems for people who have already used this option in the app and switched to `^`, we will default them back to `sh`.

PR #398, opened 2020-10-28
Adds additional option to curl codegen

webholik:
Created new PR for #251.


codegens/curl/lib/index.js
```diff
@@ -186,6 +187,14 @@ self = module.exports = {
         description: 'Set a character used to mark the continuation of a statement on the next line ' +
           '(generally, \\ for OSX/Linux, ^ for Windows)'
       },
+      {
+        name: 'Quote Type',
+        id: 'quoteType',
+        availableOptions: ['single', 'double'],
+        type: 'string',
+        default: 'single',
+        description: 'String denoting the quote type to use (single or double) for URL'
```
umeshp7, line 196: Do you think we should also specify when single and double quote are preferable like for line continuation `(generally, ...)` ? @webholik

umeshp7 (approved):

PR #400, opened 2020-10-30
Use sdk.Parse.url to for parsing urls in http, nodejs-native and python-http.client codegens.

webholik:
Removed functions used for manual parsing, added tests.


codegens/http/test/unit/converter.test.js
```diff
@@ -253,6 +253,30 @@ describe('Converter test', function () {
...
+
+    convert(request, {}, function (error, snippet) {
+      if (error) {
+        expect.fail(null, null, error);
+      }
+      expect(snippet).to.be.a('string');
+      expect(snippet).to.include('GET /{{path}}');
+      expect(snippet).to.include('Host: {{variable}}');
+    });
+  });
```
umeshp7, line 279: Also need a test with: 
```
"url": {
    "path": [
      "hello"
    ],
    "host": [
      "https://example.com/path"
    ],
    "query": [],
    "variable": []
  }
```
And have the Host: example.com as the check

codegens/http/lib/code-http-converter.js
```diff
@@ -29,18 +30,33 @@ function getOptions () {
...
-  // urlParser expects a protocal in the url
-  // If it is not present we have to add it manually
-  if (!url.match(/^([a-z][a-z0-9.+-]*:)?(\/\/)/)) {
-    url = `http://${url}`;
+
+  try {
+    url = sdk.Url.parse(request.url);
+  }
+  catch (e) {
+    url = request.url;
```
umeshp7, line 40: @webholik Can we have a unit test testing this case?

codegens/http/lib/code-http-converter.js
```diff
@@ -1,5 +1,6 @@
 let utils = require('./util'),
-  urlParser = require('url').parse;
+  _ = require('./lodash'),
+  sdk = require('postman-collection');
```
umeshp7, line 3: We'll probably have to add this to the dependencies in package.json, because these modules should be individually usable too. We usually avoid using any dependencies in these codegens to avoid increasing the size of these modules. But this is a special use case we'll have to add it.

codegens/http/test/unit/converter.test.js
webholik: That test was already added in the previous PR.

codegens/http/lib/code-http-converter.js
webholik: I added this when I was testing `URL` class in Nodejs, which throws an exception if url is malformed. But I checked `postman-collection` source and `Url.parse` doesn't throw any exception so I have removed the try catch here.

codegens/http/lib/code-http-converter.js
webholik: Done

umeshp7 (approved):

PR #418, opened 2020-11-06
Add Content-type for form-data fields if is present in the request.

webholik:
Fixes https://github.com/postmanlabs/postman-app-support/issues/8736

Support has been added for the following codegens:
* Curl
* Go
* HTTP
* Java - OkHttp
* Java - Unirest
* Libcurl
* Nodejs - Axios
* Nodejs - Native
* Powershell
* Python - http.client
* Swift
* Ocaml

The rest we couldn't support the following because there is no way to add Content-Type for these :
* Javascript - All the codegens use `FormData` api, which does not support content-type for form fields
* Nodejs - Unirest and Request
* All the PHP codegens - they use associative arrays to store formdata. Didn't see how we can add Content-Type
* Ruby - Ruby uses the HTML5 spec instead of ﻿﻿RFC 7578, and it explicitly says that formdata values cannot have Content-Type
* Csharp-Restsharp
* httpie, wget, Python-requests - They don't support formdata at all.

### DIFF
--- a/codegens/curl/lib/index.js
+++ b/codegens/curl/lib/index.js
@@ -150,7 +150,11 @@ self = module.exports = {
                 }
                 else {
                   snippet += indent + `${form('-F', format)}`;
-                  snippet += ` '${sanitize(data.key, trim)}="${sanitize(data.value, trim, true, true)}"'`;
+                  snippet += ` '${sanitize(data.key, trim)}="${sanitize(data.value, trim, true, true)}"`;
+                  if (data.contentType) {
+                    snippet += `;type=${data.contentType}`;
+                  }
+                  snippet += '\'';
                 }
               }
             });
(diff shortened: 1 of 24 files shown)


umeshp7:
@webholik do we have a newman test collection which has formdata with content-type?

umeshp7:
Also, I can see some languages we haven't added this feature. Can you elaborate in the description which codegens we have added and which ones we haven't with the reason? Will help to track this later. @webholik

webholik:
> @webholik do we have a newman test collection which has formdata with content-type?

@umeshp7 We don't have a newman test collection for it since, postman-echo ignores the content-type in form fields. I tried adding one but all the codegens passed it, since the reply from postman-echo was the same.

umeshp7:
Can you check http-bin? @webholik 
We can choose any server that respects content-type in form-data. Doesn't have to be postman-echo.

webholik:
@umeshp7 http-bin doesn't support it either

umeshp7 (approved):

PR #419, opened 2020-11-06
Use quotes for prepackagePath in deepinstall.js

### LINKED ISSUE #396: escaping space charcter in file path
I recently clone this repo and run npm install and found that one of dependency that is deepinstall.js is not there in [npmjs](https://www.npmjs.com/search?q=deepinstall.js) so dependecies should be updated.

webholik:
Fixes #396


Alphasaurs:
@webholik I have tested this change in my local repo by making changes in the mention file and it worked fine, Thanks for your quick action against this issue.

umeshp7 (approved):

Issue #435, opened 2020-12-29
Wrong URL Encode


Farhan25:
**Describe the bug**
There seems to be a bug in URL encode for PHP generator (cURL, HTTP_Request2, pecl_http) and many others languages 

**To Reproduce**
fix the URL encode

**Expected code snippet and corresponding request**
`resource_id=sc-domain%253Aexample.com` is wrong encoing
the right one is `resource_id=sc-domain%3Aexample.com` 

**Screenshots**
![image](https://user-images.githubusercontent.com/12432804/103294949-93c96380-49fb-11eb-8c21-02a1b9a3022b.png)

**Additional context**
the two URLs are different so the whole request would give two separate results.

Version of postman-code-generators/Postman app:
I have the latest version for windows v7.36.1

ShauryaAg:
The issue is that it is not replacing '%' sign correctly when it is passed as params in the Request body.

#### To Reproduce:
Pass a parameter with '%' as value to any key

#### The error:
The error is at line 19 of the [postman-code-generators/codegens/php-curl/lib/util/sanitize.js](https://github.com/postmanlabs/postman-code-generators/blob/31dc9f64295ec47b56c2912901eb39e5bdf067db/codegens/php-curl/lib/util/sanitize.js#L19)

The encodeURIComponent() function parses '%' as '%25'
![image](https://user-images.githubusercontent.com/31778302/103440145-38041200-4c69-11eb-9d1a-a2c81a5c91fa.png)

I would like to contribute to this issue.

ShauryaAg:
The issue is that it is not replacing '%' sign correctly when it is passed as params in the Request body.

#### To Reproduce:
Pass a parameter with '%' as value to any key

#### The error:
Error is at line 111 of [postman-code-generators/codegens/php-curl/lib/php-curl.js/](https://github.com/postmanlabs/postman-code-generators/blob/develop/codegens/php-curl/lib/php-curl.js#L111)

The encodeURI() function parses '%' as '%25'
![image](https://user-images.githubusercontent.com/31778302/103441940-989a4b80-4c77-11eb-810e-184388d1d263.png)


I would like to work on this issue. @abhijitkane

Farhan25:
Well I know what's the problem is, there is double encoding in there which cause the % to be encoded in the second time. 
I just mentioned it to be solved in the next update.
Thanks for your work and if It's necessary I can help you guys :)

webholik:
[...]

PR #437, opened 2021-01-02
fixed issue #435

ShauryaAg:
Replaced `encodeURI` function with `sdk.Url(urlString)` from `postman-collection` to parse the URL

### DIFF
--- a/codegens/golang/lib/index.js
+++ b/codegens/golang/lib/index.js
@@ -1,4 +1,5 @@
 var _ = require('./lodash'),
+  sdk = require('postman-collection'),
   sanitize = require('./util').sanitize,
   sanitizeMultiline = require('./util').sanitizeMultiline,
   sanitizeOptions = require('./util').sanitizeOptions,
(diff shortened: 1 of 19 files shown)


ShauryaAg:
One of test in golang is failing because the generated code snippet includes `url := "https://google.com/"` instead of `url := "https://google.com"`

Can someone help me with this? @webholik

ShauryaAg:
Made a few changes in test files as mentioned in #438

codegens/golang/lib/index.js
```diff
@@ -243,7 +244,11 @@ self = module.exports = {
     }
     codeSnippet += `${indent}"net/http"\n${indent}"io/ioutil"\n)\n\n`;
 
-    codeSnippet += `func main() {\n\n${indent}url := "${encodeURI(request.url.toString())}"\n`;
+    finalUrl = new sdk.Url(request.url.toString());
+    // URL encoding each part of Url individually
+    finalUrl = `${finalUrl.protocol}://${finalUrl.getRemote()}${finalUrl.getPathWithQuery(true)}`;
```
webholik, line 249: A url may not have protocol present. For example if the url is `postman.com/post`, then `finalURL.protocol` will return `undefined` and the finalUrl will be `undefined://postman.com/post` which is invalid. 

Infact I don't see why we need to manually construct the final URL. [url.toString()](https://github.com/postmanlabs/postman-collection/blob/cd3fa984918bf5edd6c96a6fbb2886a573439ebd/lib/collection/url.js#L197) does that for us.

webholik (changes requested):

codegens/golang/lib/index.js
ShauryaAg: Ah, sorry. I was unfamiliar with `url.toString()`. I will make a necessary changes as soon as possible.

ShauryaAg:
@webholik I have made the necessary changes, however `url.toString()` method doesn't URL encode `'` (single-quote) and `"` (double-quote) character.

ShauryaAg:
@webholik the test fails as it doesn't URL encode `'` and `"` character. 

I think a better approach would be to parse URLs using `postman-collection` only but then put checks if **protocol** is empty (and any other conditions that I might have missed)

Something like
```finalUrl = `${url.protocol ? `${url.protocol}://` : '' }${url.getRemote()}${finalUrl.getPathWithQuery(true)}` ```

Issue #447, opened 2021-01-22
Add a Google App Script (UrlFetchApp) code generator


rayliuca:
**Is your feature request related to a problem? Please describe.**
No code generator is compatible with Google Apps Script and GAS doesn't support regular JS libraries regarding API requests

**Describe the solution you'd like**
Implement postman code generation for UrlFetchApp

**Additional context**
An issue on the postman-app-support:
https://github.com/postmanlabs/postman-app-support/issues/3298

My implementation:
https://github.com/rayliuca/postman-code-generators/tree/develop/codegens/js-urlfetchapp

My implementation is based on the javascript-xhr code generator, but do note that the UrlFetchApp class is non-existent on node.js so it is impossible to run the newman tests. I adapted the unit tests from the xhr variant and manually tested the generated codes on GAS to make sure it has most of the common functionalities.

Snippets:
https://github.com/rayliuca/postman-code-generators/tree/develop/codegens/js-urlfetchapp/unitTestSnippets

PR #449, opened 2021-01-22
Add support for Google App Script (UrlFetchApp)

rayliuca:
Added support for Google Apps Script's UrlFetchApp class as per #447 

Google documentation:
https://developers.google.com/apps-script/reference/url-fetch/url-fetch-app


rayliuca:
My implementation is based on the javascript-xhr code generator, but do note that the UrlFetchApp class is non-existent on Node.js so it is impossible to run the Newman tests. I adapted the unit tests from the xhr variant and manually tested the generated codes on GAS to make sure it has most of the common functionalities.

Sample response of the [post_formData.js](https://github.com/rayliuca/postman-code-generators/blob/develop/codegens/js-urlfetchapp/unitTestSnippets/post_formData.js)

image

clairefro:
Would love to see this feature added!

rsam036 (approved):

rsam036 (approved):

Issue #351, opened 2020-10-05
Feature request : Add react code in generate code snippet. 


Alphasaurs:
It would be great to get a direct react snippet to link in our project.

2019UCP1350:
I would like to build this. Can I work on it?

vickOnRails:
It would be a great idea to get snippet generation for React since most client apps are built with React.

@2019UCP1350 what's up? 

Does anyone know what it would take to get this working?

2019UCP1350:
@vickOnRails I had almost completed this code snippet and just writing some test to check it. You can take some idea from existing snippet to understand how it's working.

vickOnRails:
Oh great @2019UCP1350 

Do you have a repo or something I can look at?

2019UCP1350:
Here is the link to the pull request https://github.com/postmanlabs/postman-code-generators/pull/473.
You can take a look at codegen here.

PR #473, opened 2021-03-27
Created  Code Generator for React JS  in reference to issue #351

2019UCP1350:
As the popularity of react js to build websites and web apps is increasing, therefore, there is a need for the hour to build codegen for reactjs.

 Here are the few websites which show the popularity of reactjs 
 https://ezeelive.com/reactjs-popularity/
 https://www.heliossolutions.co/blog/top-7-reasons-why-reactjs-development-is-so-popular/

Moreover, One of the major social media web-app Whatsapp is also built-in reactjs


Sharath-Postman:
@abhijitkane if you could look into this.

original ref issue -- https://github.com/postmanlabs/gsoc/issues/36

2019UCP1350:
> @abhijitkane 
Could you have a look at my work?

umeshp7:
@2019UCP1350 We already have [nodejs-axios](https://github.com/postmanlabs/postman-code-generators/tree/develop/codegens/nodejs-axios). 
Could you summarize in a comment how different is this codegen from the that one?

VShingala:
@2019UCP1350 As @umeshp7 suggested, we already have added support for reactjs (axios) framework. From what I gathered in PR, it also do it in axios right?

In that case, can you mention if existing code gen for reactjs works for you as expected and if not does this PR add any new features to it?

2019UCP1350:
> @2019UCP1350 As @umeshp7 suggested, we already have added support for nodejs (axios) framework. From what I gathered in PR, it also do it in axios right?
> 
> In that case, can you mention if existing code gen for axios works for you as expected and if not does this PR add any new features to it?

When I had made this PR, Someone has requested the codegen snippet for reactjs, and it was not there.
Moreover, I used the try, catch and async, await syntax rather than nested then and catch syntax making the code snippet look pleasant and cleaner.
Also, currently, only node js Axios support is there, and snippet generates code for node js rather than react js which uses import differently than node js.

VShingala:
@2019UCP1350 Makes sense! Thanks for the contribution and fast resonse. We'll evaluate these changes and see if we can accommodate them in Postman app codegen snippets.

dhwaneetbhatt (changes requested):
@2019UCP1350 Rather than this being a different option altogether, I think passing it as an option to the original NodeJS-Axios generator would be good. We can keep it multi-select to be NodeJS or ReactJS, and it can generate any additional different code for ReactJS. Most of the code would remain the same. What are your thoughts on this?

We have recently upgraded the NodeJS-Axios generator with changes to produce async/await compatible code. Could you please update your fork from the latest upstream develop and see if this can work as part of the same generator?

PR #477, opened 2021-04-12
Add proper formatting of JSON in Dart and C#

### LINKED ISSUE #459: C# use verbatim string
Current generated requests are a tragedy to read. We could improve this a bit by introducing a verbatim string with @ prefix. Instead of this:

`request.AddParameter("application/json", "{  \r\n  \"order\":{  \r\n      \"items\":[  \r\n         {  \r\n            \"name\": ...`

It could be like this:

```
request.AddParameter("application/json", @"{
    ""order"": {
        ""items"": [
            {
                ""name"": ...
```

If accepted, it would probably be a good idea to make this as a toggleable setting, for users that want to have these long one-lines or for others that want to have it in semi-formatted way.

A good source for information - https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/tokens/verbatim

webholik:
Fixes #459 
C#:
```
using RestSharp;
namespace HelloWorldApplication {
  class HelloWorld {
    static void Main(string[] args) {
      var client = new RestClient("http://example.com");
      client.Timeout = -1;
      var request = new RestRequest(Method.GET);
      request.AddHeader("Content-Type", "application/json");
      var body = @"{" + "\n" +
      @"    ""hello"": ""world""," + "\n" +
      @"    ""this"": true," + "\n" +
      @"    ""call"": [1,2,3,null]" + "\n" +
      @"}";
      request.AddParameter("application/json", body,  ParameterType.RequestBody);
      IRestResponse response = client.Execute(request);
      Console.WriteLine(response.Content);
    }
  }
}
```

Dart:
```
import 'dart:convert';
import 'package:http/http.dart' as http;

void main() async {
  var headers = {
    'Content-Type': 'application/json'
  };
  var request = http.Request('GET', Uri.parse('http://example.com'));
  request.body = json.encode({
      "hello": "world",
      "this": true,
      "call": [
          1,
          2,
          3,
          null
      ]
  });
  request.headers.addAll(headers);
  
  http.StreamedResponse response = await request.send();
  
  if (response.statusCode == 200) {
    print(await response.stream.bytesToString());
  }
  else {
    print(response.reasonPhrase);
  }
  
}
```

### DIFF
--- a/codegens/dart-http/lib/index.js
+++ b/codegens/dart-http/lib/index.js
@@ -30,8 +30,19 @@ function parseUrlEncoded (body, indent, trim) {
  *
  * @param {Object} body Raw body data
  * @param {Boolean} trim indicates whether to trim string or not
+ * @param {String} contentType the content-type of request body
  */
-function parseRawBody (body, trim) {
+function parseRawBody (body, trim, contentType) {
+  if (contentType && (contentType === 'application/json' || contentType.match(/\+json$/))) {
+    try {
+      let jsonBody = JSON.parse(body);
+      return `request.body = json.encode(${JSON.stringify(jsonBody, null, 4)});`;
+
+    }
+    catch (error) {
+      // Do nothing
+    }
+  }
   return `request.body = '''${sanitize(body, trim)}''';`;
 }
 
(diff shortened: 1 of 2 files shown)


umeshp7 (approved):

codegens/dart-http/lib/index.js
```diff
@@ -30,8 +30,19 @@ function parseUrlEncoded (body, indent, trim) {
...
  * @param {Object} body Raw body data
  * @param {Boolean} trim indicates whether to trim string or not
+ * @param {String} contentType the content-type of request body
  */
-function parseRawBody (body, trim) {
+function parseRawBody (body, trim, contentType) {
+  if (contentType && (contentType === 'application/json' || contentType.match(/\+json$/))) {
+    try {
+      let jsonBody = JSON.parse(body);
+      return `request.body = json.encode(${JSON.stringify(jsonBody, null, 4)});`;
```
umeshp7, line 39: The space will depend on the indentation right? @webholik

umeshp7 (changes requested):

codegens/dart-http/lib/index.js
webholik: @umeshp7 The problem is that if the indentation is using tabs, then we have no way to handle that with `JSON.stringify`. So I decided to use constant indentation for all cases.

umeshp7 (approved):

PR #479, opened 2021-04-16
Add proper handling for query and variables of graphql body

umeshp7:
Add proper handling for query and variables of graphql body


umeshp7:
@shreys7 Any reason for JSON.parse-ing the graphql.variables and then JSON.stringify-ing again ? What were the issues of using the graphql.variables string as is?

PR #507, opened 2021-07-15
fix: use url.toString()

### LINKED ISSUE #506: Failing url encoding in multiple codegens due to back call to Url.getQueryString
`Url.getQueryString({encode: true})` is no longer supported (https://github.com/postmanlabs/postman-collection/pull/1198/files#diff-0267b663041b4094bb4802be9641a818f458a54127910a9126e61ae9d499232eL313). 

This leads to urls that are not properly encoded in the following codegens:

- curl
- swift
- java-unirest

https://github.com/postmanlabs/postman-code-generators/search?q=getQueryString

jpoehnelt:
closes #506 and removes manual construction of url

### DIFF
--- a/codegens/curl/lib/util.js
+++ b/codegens/curl/lib/util.js
@@ -139,32 +139,7 @@ var self = module.exports = {
     if (!urlObject) {
       return url;
     }
-    if (urlObject.protocol) {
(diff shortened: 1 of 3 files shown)


jpoehnelt:
will repopen with proper fixes :)

PR #509, opened 2021-07-15
ci: replace dummyFile with fakeFile

jpoehnelt:
I believe fake is more exact language for this usage.  related change to gitignore in #508


webholik:
@jpoehnelt Thanks for the PR but I see no problem with the current naming. So closing this.

PR #515, opened 2021-08-03
Remove postman-collection dependency in codegens

webholik:


umeshp7:
@webholik Are we still working on this?

webholik:
@umeshp7 Nope. I don't remember why I stopped working on this.  Will take a re-look to decide whether it is still relevant.

PR #536, opened 2021-10-12
Fix the usage of -X cli option in curl codegen [WIP]

umeshp7:


webholik:
No longer needed due to #585

PR #541, opened 2021-10-24
Add an option to remove new line characters from raw body in cURL cod…

### LINKED ISSUE #540: Add an option to remove new line characters from body in cURL code snippet
When generating cURL code snippet, even though the "Generate multiline snippet" option is disabled, the code snipped can became multiline when raw body ex. JSON contains new line character(s). It's especially annoying when copy-pasting in old Windows terminals that don't handle it properly. By adding this option we could have neatly formatted JSON in body editor area and one line cURL code snippet.

A separate option in "code generation settings" to strip any new line characters from body (disabled by default). When enabled it would replace any "\n" with "".

**Additional context**
PR #541

Pawloland:
…e snippet

Fixes #540


umeshp7:
@Pawloland we already have an option regarding multiLine. Can you see if this can be implemented along with that? Or do you see a use case keeping them separate? Happy to discuss.

PR #585, opened 2022-02-26
Do not add -X option in curl if -L is present and method is POST.

webholik:
Fixes postmanlabs/postman-app-support#10581

### DIFF
--- a/codegens/curl/lib/index.js
+++ b/codegens/curl/lib/index.js
@@ -13,7 +13,7 @@ self = module.exports = {
     if (!_.isFunction(callback)) {
       throw new Error('Curl-Converter: callback is not valid function');
     }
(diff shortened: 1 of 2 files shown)


codegens/curl/lib/index.js
```diff
@@ -47,6 +47,21 @@ self = module.exports = {
     if (request.method === 'HEAD') {
       snippet += ` ${form('-I', format)} ${quoteType + url + quoteType}`;
     }
+    else if (request.method === 'POST' &&
```
umeshp7, line 50: Let's add other cases where -X and -L will cause issues when used together

dhwaneetbhatt:
Fixed in https://github.com/postmanlabs/postman-code-generators/pull/666, closing this PR.

PR #599, opened 2022-03-25
Add warning comments before misleading code in JS-xhr

webholik:
If a user tries send a GET request with a body or with the `Cookie` header set, the add a comment warning them that browser will strip them away before sending the request. For example:
```javascript
// WARNING: For GET requests, body is set to null by browsers.
var data = JSON.stringify({
  'hello': 'world'
});

var xhr = new XMLHttpRequest();
xhr.withCredentials = true;

xhr.addEventListener("readystatechange", function() {
  if(this.readyState === 4) {
    console.log(this.responseText);
  }
});

xhr.open("GET", "https://postman-echo.com/get");
// WARNING: Cookies will be stripped away by the browser before sending the request.
xhr.setRequestHeader("Cookie", "cookie");
xhr.setRequestHeader("Content-Type", "application/json");

xhr.send(data);
```

### DIFF
--- a/codegens/js-xhr/lib/index.js
+++ b/codegens/js-xhr/lib/index.js
@@ -150,6 +150,9 @@ function parseHeaders (headers) {
   if (!_.isEmpty(headers)) {
     headers = _.reject(headers, 'disabled');
     _.forEach(headers, function (header) {
+      if (_.capitalize(header.key) === 'Cookie') {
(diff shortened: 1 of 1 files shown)


umeshp7 (approved):

umeshp7:
Looks good. @webholik 
Just wondering does it makes sense to even generate a snippet like this. But we can't take the decision for the user and simply generate a POST request. Should we suggest using POST method along with the warning you have added?

PR #651, opened 2022-11-02
Fix/rest sharp code gen

LuisTejedaS:
Fix generated code for Restsharp C# library


Carlos-Veloz (approved):

erfemega (approved):
Great Job

umeshp7:
@alexeyzimarev Request your review here

codegens/csharp-restsharp/lib/restsharp.js
```diff
@@ -158,6 +225,14 @@ self = module.exports = {
...
         description: 'Remove white space and additional lines that may affect the server\'s response'
+      },
+      {
+        name: 'Set communication type',
+        id: 'asyncType',
+        type: 'enum',
+        availableOptions: ['async', 'sync'],
+        default: 'async',
+        description: 'Set if the requests will be asynchronous or synchronous'
       }
```
umeshp7, line 236: {
        name: 'Execution type',
        id: 'execType',
        type: 'enum',
        availableOptions: ['async', 'sync'],
        default: 'async',
        description: 'Choose if the requests will be executed asynchronously or synchronously'
      }

umeshp7 (commented):
I wanted to understand if there is any absolute requirement of adding async and sync option? 

Are there any cases we are missing if we don't include this?

alexeyzimarev:
My main question is, if we assume the generated code is correct, how can it prevent people from instantiating the client for each request? Maybe it's not an issue, and adding a comment about it would suffice...

alexeyzimarev:
I think I will be able to handle the instantiation in RestSharp by the next iteration.

alexeyzimarev:
> @alexeyzimarev Request your review here

For some reason, I don't see the review approval button, although I am in the list...

codegens/csharp-restsharp/lib/restsharp.js
umeshp7: @alexeyzimarev Thank you.
@LuisTejedaS Can we remove this option altogether?

LuisTejedaS: Hi @umeshp7 i removed the option and the code regarding sync, we are only using async

alexeyzimarev: 👍 Sync calls are just extensions using the AsyncHelper wrapper. It's a workaround, and people should only be using async calls anyway.

umeshp7 (approved):

PR #664, opened 2023-01-30
Removed semaphore from swift snippet

SahilChoudhary22:
## Fixes
- [#10053](https://github.com/postmanlabs/postman-app-support/issues/10053) - Removed semaphore from the generated swift language snippet
- Added `pull_request` as GitHub actions workflow trigger
- Removed unused `followRedirect` option from the swift language codegen


codegens/swift/lib/swift.js
```diff
@@ -250,11 +250,11 @@ self = module.exports = {
         description: 'Remove white space and additional lines that may affect the server\'s response'
       },
       {
-        name: 'Follow redirects',
```
akshaydeo, line 253: @SahilChoudhary22 any specific reason we removed this option? We might need to block the redirects if its unchecked right? Or you removed as it was not getting used.

cc @VShingala do we not block redirects if this flag is off 🙇‍♂️

SahilChoudhary22: @akshaydeo Yes, we observed that this option was not getting used in the business logic and therefore it being `true` or `false` was not making any difference to the output.

VShingala: @akshaydeo Swift by default follows redirection and from our research, there was no easy way to control this behavior from swift using some option. So yeah as it's not possible from Swift language itself, the option was not used.

webholik (approved):

PR #666, opened 2023-02-02
Do not add HTTP method explicitly in curl when not required

dhwaneetbhatt:
- For `HEAD`, do not add `--request`
- For `GET` only add if it has a request body and the request has body pruning is not enabled
- For `POST`, only add if there is no request body
- Add it explicitly in cases where the user has selected to follow redirects and also retain the original HTTP method while doing it
- Introduced a new option, `followOriginalHttpMethod` with the default value `false`
- We're trying to make the experience for users better by also honouring request settings as it is sometimes frustrating to tweak snippet settings in addition to request settings.
  - `followOriginalHttpMethod` is also an option in request in Postman. To better enable integration with the request editor, we honour the value if present in request and fallback to options if not present.
  - 'followRedirect` also honours to request level `followRedirects` option is present, falling back to options if not present.

Fixes https://github.com/postmanlabs/postman-app-support/issues/10581.

More here: https://daniel.haxx.se/blog/2015/09/11/unnecessary-use-of-curl-x/

### DIFF
--- a/codegens/curl/lib/index.js
+++ b/codegens/curl/lib/index.js
@@ -244,7 +244,7 @@ self = module.exports = {
         id: 'followOriginalHttpMethod',
         type: 'boolean',
         default: false,
-        description: 'Use the original HTTP method when following redirects'
+        description: 'Redirect with the original HTTP method instead of the default behavior of redirecting with GET'
       },
       {
         name: 'Trim request body fields',
--- a/codegens/curl/lib/index.js
+++ b/codegens/curl/lib/index.js
@@ -3,6 +3,7 @@ var sanitize = require('./util').sanitize,
   getUrlStringfromUrlObject = require('./util').getUrlStringfromUrlObject,
   addFormParam = require('./util').addFormParam,
   form = require('./util').form,
+  shouldAddHttpMethod = require('./util').shouldAddHttpMethod,
   _ = require('./lodash'),
   self;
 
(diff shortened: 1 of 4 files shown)


codegens/curl/lib/index.js
```diff
@@ -236,6 +239,13 @@ self = module.exports = {
         default: true,
         description: 'Automatically follow HTTP redirects'
       },
+      {
+        name: 'Follow original HTTP method',
+        id: 'followOriginalHttpMethod',
+        type: 'boolean',
+        default: false,
+        description: 'Use the original HTTP method when following redirects'
```
VShingala, line 247: image
I think we can keep the name and description the same as the request setting option.

codegens/curl/lib/util.js
```diff
@@ -199,5 +201,77 @@ var self = module.exports = {
...
+   * Decide whether we should add the HTTP method explicitly to the cURL command.
+   *
+   * @param {Object} request
+   * @param {Object} options
+   *
+   * @returns {Boolean}
+   */
+  shouldAddHttpMethod: function (request, options) {
+    const followOriginalHttpMethod =
+      _.get(request, 'protocolProfileBehavior.followOriginalHttpMethod', options.followOriginalHttpMethod),
```
VShingala, line 248: @dhwaneetbhatt Not sure here, Shouldn't option override the behaviour of what's present in request if present?

codegens/curl/lib/util.js
```diff
@@ -199,5 +201,77 @@ var self = module.exports = {
...
+      return true;
+    }
+
+    if (body.isEmpty()) {
+      return true;
+    }
+
+    if (body.mode === 'formdata' || body.mode === 'urlencoded') {
+      let memberCount = 0;
+      body[body.mode].members && body[body.mode].members.forEach((param) => {
```
VShingala, line 226: Add this extra condition `body[body.mode] &&` at start to avoid typeerrors in future.

codegens/curl/lib/util.js
dhwaneetbhatt: @VShingala The default value of the option is always `false`. There is no undefined state, so we cannot determine if the user or the default value sets the option's current value.

I can change this to `_.get(request, 'protocolProfileBehavior.followOriginalHttpMethod', false) || options.followOriginalHttpMethod` to honour the option too. But it won't be an "override". Any better ideas?

VShingala: So my main intention was to clarify whether `protocolProfileBehavior.followOriginalHttpMethod` should take precedence or `options.followOriginalHttpMethod`. Current _.get statement is fine, just what should be first followed is something we can discuss.

dhwaneetbhatt: As per the discussion, the snippet-level option overrides the request-level option. Let me re-think the condition here. Thanks for initiating this discussion.

PR #672, opened 2023-02-11
Update util.js

lucioerlan:
🔨🔨 adding Performance: the refactored code might perform a little better than the original code

- Clarity: The refactored code is clearer and easier to read, since the type logic is encapsulated in separate objects.
- Reusability: The refactored code allows reusing the type checker functions in other parts of the code.
- Simplicity: The refactored code is simpler, as type checks are performed in separate functions, making the main code shorter and more concise.
- Maintainability: The refactored code is easier to maintain, as future modifications to the type checks can be made in the separate functions, without affecting the main code.
- Performance: The refactored code may have slightly better performance than the original code, as the type checker functions are executed only once, instead of repeating the check logic for each option.

### DIFF
--- a/codegens/nodejs-axios/lib/util.js
+++ b/codegens/nodejs-axios/lib/util.js
@@ -11,6 +11,7 @@ function sanitize (inputString, trim) {
   if (typeof inputString !== 'string') {
     return '';
   }
+
   (trim) && (inputString = inputString.trim());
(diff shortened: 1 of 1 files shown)


dhwaneetbhatt:
@lucioerlan could you please update your fork with upstream changes from develop? We have fixed the pipeline. It would allow us to run tests on your changes and move ahead for review.

PR #685, opened 2023-03-20
Released v1.6.0.

VShingala:


dhwaneetbhatt (approved):

VShingala:
There seems to be some issue while installing dependencies for tests. As this PR only bumps version, tests should not be affected so merging it.

PR #694, opened 2023-03-29
Release v1.7.1

dhwaneetbhatt:
Adding language labels for new languages.


webholik (approved):

webholik:
@dhwaneetbhatt Can we add a test case for this, so we don't miss it in the future?

dhwaneetbhatt:
> @dhwaneetbhatt Can we add a test case for this, so we don't miss it in the future?
Added

webholik (approved):

PR #710, opened 2023-06-12
[FEATURE] Codegen Dart Dio Support

aman-v-singh:
- Language Support for Dart Dio variant


codegens/dart-dio/lib/lodash.js
dhwaneetbhatt, line 1: We can use `lodash` directly without adding a copy. Check rust codegen.

codegens/dart-dio/npm-shrinkwrap.json
dhwaneetbhatt, line 1: Can use `package-lock.json` instead of `npm-shrinkwrap.json`. Check rust codegen.

test/codegen/newman/fixtures/formdataFileCollection.json
dhwaneetbhatt, line 1: Do not commit this file as it will keep changing.

dhwaneetbhatt (changes requested):

npm/ci-requirements.sh
```diff
@@ -80,13 +80,30 @@ echo "Installing curl"
...
-  http: ^0.12.2
+  http: ^1.0.0
+''' > pubspec.yaml
+  dart pub get
+popd &>/dev/null;
+
+echo "Installing dependencies required for tests in codegens/dart-dio"
+pushd ./codegens/dart-dio &>/dev/null;
+  wget -q https://storage.googleapis.com/dart-archive/channels/stable/release/latest/linux_packages/dart_3.0.4-1_amd64.deb
+  sudo dpkg -i dart_3.0.4-1_amd64.deb
```
dhwaneetbhatt, line 100: We are doing this two times, which is not required. Dart installation can be done only once, and then we can build two different `pubspec.yaml` files based on the variations.

codegens/dart-dio/package.json
```diff
@@ -0,0 +1,36 @@
...
+  },
+  "author": "Postman Labs <help@getpostman.com>",
+  "license": "Apache-2.0",
+  "homepage": "https://github.com/postmanlabs/code-generators/tree/master/codegens/dart-dio",
+  "dependencies": {
+    "lodash": "4.17.21"
+  },
+  "devDependencies": {},
+  "engines": {
+    "node": ">=8"
```
dhwaneetbhatt, line 34: Let's keep this Node 12. We've updated the main library's min version to be Node 12.

dhwaneetbhatt (approved):

PR #714, opened 2023-07-05
[GITHUB-10139][Fix] Generated Swift code not working for multipart/form-data video upload

aman-v-singh:
### Context
  - In postman-app-support, [Issue #10139](https://github.com/postmanlabs/postman-app-support/issues/10139) The user has described an issue they were facing with `Swift` code generated for `Content-Type` `multipart/form-data`.

### Root-Cause-Analysis
  - When the user was using the generated Swift code for uploading a `.MOV` video file, they were facing an issue on reaching this line:-
  `let fileContent = String(data: fileData, encoding: .utf8)!`
  
  - On analysis, it is found that the error is arising because the generated code is trying to encode video data in `UTF-8` format to be converted into `String` type, but we must realise every character in `fileData` would not be encodable in `UTF-8` format.

### Solution
  - Instead of converting file data to `String` type, we will change the `body` variable from `String` to `Data` type. This way we don't have to convert the `fileContent` to `String`.
  
  - For details refer to this [answer](https://stackoverflow.com/questions/68717871/sending-mov-video-file-as-part-of-post-request-to-own-server-as-part-of-a-multi).
  
### Code Generated

```swift
import Foundation
#if canImport(FoundationNetworking)
import FoundationNetworking
#endif

let parameters = [
    [
        "key": "VideoFile",
        "src": "{VIDEO_FILE_PATH}",
        "type": "file"
    ]] as [[String: Any]]

let boundary = "Boundary-\(UUID().uuidString)"
var body = Data()
var error: Error? = nil
for param in parameters {
    if param["disabled"] != nil { continue }
    let paramName = param["key"]!
    body += Data("--\(boundary)\r\n".utf8)
    body += Data("Content-Disposition:form-data; name=\"\(paramName)\"".utf8)
    if param["contentType"] != nil {
        body += Data("\r\nContent-Type: \(param["contentType"] as! String)".utf8)
    }
    let paramType = param["type"] as! String
    if paramType == "text" {
        let paramValue = param["value"] as! String
        body += Data("\r\n\r\n\(paramValue)\r\n".utf8)
    } else {
        let paramSrc = param["src"] as! String
        let fileURL = URL(fileURLWithPath: paramSrc)
        if let fileContent = try? Data(contentsOf: fileURL) {
            body += Data("; filename=\"\(paramSrc)\"\r\n".utf8)
            body += Data("Content-Type: \"content-type header\"\r\n".utf8)
            body += Data("\r\n".utf8)
            body += fileContent
            body += Data("\r\n".utf8)
        }
    }
}
body += Data("--\(boundary)--\r\n".utf8);
let postData = body


var request = URLRequest(url: URL(string: "https://postman-echo.com/post")!,timeoutInterval: Double.infinity)
request.addValue("multipart/form-data; boundary=\(boundary)", forHTTPHeaderField: "Content-Type")

request.httpMethod = "POST"
request.httpBody = postData

    guard let data = data else {
        print(String(describing: error))
        exit(EXIT_SUCCESS)
    }
    print(String(data: data, encoding: .utf8)!)
    exit(EXIT_SUCCESS)
}

task.resume()
dispatchMain()
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ## [Unreleased]
 
+- Fix for - [#10139](https://github.com/postmanlabs/postman-app-support/issues/10139) Modify Swift codegen to work with multipart/form-data format, used for video file upload
+
 ## [v1.8.0] - 2023-06-27
 
 -   Fix for - [#10521](https://github.com/postmanlabs/postman-app-support/issues/10521) Add support for Dart Dio snippet generation
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,9 @@
 
 ## [Unreleased]
 
-- Fix for - [#10139](https://github.com/postmanlabs/postman-app-support/issues/10139) Modify Swift codegen to work with multipart/form-data format, used for video file upload
+### Fixed
+
+-   Fix for - [#10139](https://github.com/postmanlabs/postman-app-support/issues/10139) Modify Swift codegen to work with multipart/form-data format, used for video file upload
 
 ## [v1.8.0] - 2023-06-27
 
--- a/codegens/swift/lib/swift.js
+++ b/codegens/swift/lib/swift.js
@@ -128,8 +128,8 @@ function parseFormData (body, mode, trim, indent) {
   bodySnippet += `${indent}} else {\n`;
   bodySnippet += `${indent.repeat(2)}let paramSrc = param["src"] as! String\n`;
   bodySnippet += `${indent.repeat(2)}let fileURL = URL(fileURLWithPath: paramSrc)\n`;
-  bodySnippet += `${indent.repeat(2)}if let fileContent = try? Data(contentsOf: fileURL) {`;
-  bodySnippet += `${indent.repeat(3)}body += ("; filename=\\"\\(paramSrc)\\"\\r\\n".utf8)\n`;
+  bodySnippet += `${indent.repeat(2)}if let fileContent = try? Data(contentsOf: fileURL) {\n`;
+  bodySnippet += `${indent.repeat(3)}body += Data("; filename=\\"\\(paramSrc)\\"\\r\\n".utf8)\n`;
   bodySnippet += `${indent.repeat(3)}body += Data("Content-Type: \\"content-type header\\"\\r\\n".utf8)\n`;
   bodySnippet += `${indent.repeat(3)}body += Data("\\r\\n".utf8)\n`;
   bodySnippet += `${indent.repeat(3)}body += fileContent\n`;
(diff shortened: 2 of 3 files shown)


VShingala (commented):
@aman-v-singh Have we tested that this change doesn't affect existing flow. i.e. Other files which are in text format can sill be sent correctly in form-data body?

VShingala:
@aman-v-singh Could you also attach updated snippet here in description once? Also let's communicate this latest snippet with the User that reported this issue and make sure it works for them to get their feedback.

aman-v-singh:
@VShingala I have added the newly generated code above in the description and notified the user of the same.

> @aman-v-singh Have we tested that this change doesn't affect existing flow. i.e. Other files which are in text format can sill be sent correctly in form-data body?

The Newman tests take care of these different possibilities.
eg:- `Single/multiple file upload via form-data` and `Binary file upload via form-data` check for different types of files. 
- https://github.com/postmanlabs/postman-code-generators/blob/develop/test/codegen/newman/fixtures/formdataFileCollection.json
- Link to Unit test output:- https://github.com/postmanlabs/postman-code-generators/actions/runs/5475200225/jobs/9970978873?pr=714#step:7:23135

VShingala:
@aman-v-singh There is a bit of catch with `forma-data` type of bodies and how files are stored in collection if selected. For example, the collection that you linked contains path to a file which is for the user that created this collection. This file doesn't exist in pipelines system or repository anywhere. As such actual data within files is not being sent with the request.

Which is why just wanted to make sure this is working correctly, could you give it a try manually and verify once?

image

dhwaneetbhatt:
> 



> @aman-v-singh There is a bit of catch with `forma-data` type of bodies and how files are stored in collection if selected. For example, the collection that you linked contains path to a file which is for the user that created this collection. This file doesn't exist in pipelines system or repository anywhere. As such actual data within files is not being sent with the request.
> 
> Which is why just wanted to make sure this is working correctly, could you give it a try manually and verify once?
> 
> image

@VShingala these files and paths are generated on the fly when test case runs and this path gets updated locally. We usually don't commit the changes because it depends on the machine being run.

VShingala:
@dhwaneetbhatt Thanks! Checked this out and Didn't know about this till now. This was a practice of habit for me while I was working on this as this file manipulation did not exist. 😅

For ref, below is code responsible for manipulation.

- https://github.com/postmanlabs/postman-code-generators/blob/989648dae5a792df518297400cd284185ac42b6b/npm/test.sh#L16
- https://github.com/postmanlabs/postman-code-generators/blob/32e76dd2343b94b503af60a45ca545f1da9992ff/npm/addPathToFormdataFile.js#L7

VShingala (approved):

PR #720, opened 2023-08-04
Swift Async

maxxfrazer:
- Added async option for Swift code snippet options.
  - Default as `false`, traditional completion handler style.
- Changed EXIT_SUCCESS on existing version to EXIT_FAILURE where it actually fails.

The async code looks like:

```swift
do {
    let (data, response) = try await URLSession.shared.data(for: request)
    print(String(data: data, encoding: .utf8)!)
    exit(EXIT_SUCCESS)
} catch {
    print("A URL error occurred: \(error)")
    exit(EXIT_FAILURE)
}
```

Current non-async code:

```swift
    guard let data = data else {
        print(String(describing: error))
        exit(EXIT_FAILURE)
    }
    print(String(data: data, encoding: .utf8)!)
    exit(EXIT_SUCCESS)
}
```

IMO the async method should be default, but wanted to get your feedback first!

> I have noted in ADDITIONAL_DEPENDENCIES.md that the swift version should be a minimum of 5.5 (released Sept 2021)

### DIFF
--- a/ADDITIONAL_DEPENDENCIES.md
+++ b/ADDITIONAL_DEPENDENCIES.md
@@ -37,7 +37,7 @@
   1. [wget](https://www.gnu.org/software/wget)
 
 ### swift 
-  1. [swift-5.0.1](https://swift.org/builds/swift-5.0.1-release/ubuntu1604/swift-5.0.1-RELEASE)
+  1. [swift-5.5](https://download.swift.org/swift-5.5-release/ubuntu1604/swift-5.5-RELEASE)
   2. clang-3.6
   3. libicu-dev
   4. libpython2.7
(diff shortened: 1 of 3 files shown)


maxxfrazer:
@dhwaneetbhatt @SahilChoudhary22 you helped me with my last Swift codegen update, hopefully you see this as an improvement too!

maxxfrazer:
TIL `await URLSession.shared.data` isn't available on linux? will take a deeper dive.

maxxfrazer:
Going to skip this PR. Unfortunately async isn’t available on Linux exactly the same way as it is for iOS/macOS etc.
I think a huge majority of Swift developers would benefit from it, but unfortunately not _all_ of them.

Issue #719, opened 2023-08-02
golang : ioutil.ReadAll is deprecated from Go 1.16


panapol-p:
I saw deprecated function in go code generator

after Go 1.16 `ioutil.ReadAll(...)` is deprecated, we should use `io.ReadAll(...)` instead

Doc ref : https://pkg.go.dev/io/ioutil#ReadAll
Code : https://github.com/postmanlabs/postman-code-generators/blob/2411e41a03c59515fb78e363b0c93ba57f7d030b/codegens/golang/lib/index.js#L300C82-L300C96

panapol-p:
Hi, @theonly1me 
could you pleaes help to review this change.

panapol-p:
merged #721 , #762 

@theonly1me  and @VShingala 
Thank you for the review and the commit fixes.

PR #721, opened 2023-08-06
fix(golang)/use io instead of io/ioutil

panapol-p:
I saw deprecated function in go code generator

after Go 1.16 ioutil.ReadAll(...) is deprecated, we should use io.ReadAll(...) instead

Doc ref : https://pkg.go.dev/io/ioutil#ReadAll
issue : #719


theonly1me (approved):
Thanks for your contribution @panapol-p!

The changes look good to me, @VShingala, please feel free to take a look once.

VShingala (approved):

PR #734, opened 2024-04-17
Escape special characters in codegen

aman-v-singh:
- Escape special characters to preserve their literal meaning within double quotes

### DIFF
--- a/codegens/powershell-restmethod/lib/index.js
+++ b/codegens/powershell-restmethod/lib/index.js
@@ -293,8 +293,8 @@ function convert (request, options, callback) {
                         `${request.method}' -Headers $headers`;
   }
   else {
-    codeSnippet += `$response = Invoke-RestMethod '${request.url.toString()}' -CustomMethod ` +
-                        `'${request.method}' -Headers $headers`;
+    codeSnippet += `$response = Invoke-RestMethod '${request.url.toString().replace(/'/g, '\'\'')}' -CustomMethod ` +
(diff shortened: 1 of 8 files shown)


VShingala:
@aman-v-singh Tests are failing. Can you fix them and request review?

VShingala (changes requested):
@aman-v-singh Can we add unit/newman tests cases related to this change?

VShingala (changes requested):
@aman-v-singh Since previous fix that we did also addresses other coddegens (see list in below PR), can we make sure that changes made to `sanitize` function is also addressed in other codegens?

https://github.com/postmanlabs/postman-code-generators/pull/289

codegens/powershell-restmethod/lib/index.js
```diff
@@ -293,8 +293,8 @@ function convert (request, options, callback) {
                         `${request.method}' -Headers $headers`;
   }
   else {
-    codeSnippet += `$response = Invoke-RestMethod '${request.url.toString()}' -CustomMethod ` +
-                        `'${request.method}' -Headers $headers`;
+    codeSnippet += `$response = Invoke-RestMethod '${request.url.toString().replace(/'/g, '\'\'')}' -CustomMethod ` +
```
VShingala, line 296: Let's separate out this .replace logic to a common function and utilise it instead of direct logic.

VShingala (changes requested):
We're missing unit test for powershell changes. Let's also add test for it.

VShingala (approved):

PR #735, opened 2024-04-30
[FIX] cURL code snippet contains encoded value for special char + instead of char itself

aman-v-singh:
## Overview

This PR fixes issue where cURL snipet generated did not follow same behaviour as Postman Send request flow. i.e. `+` character in query were being encoded, but ideally they should not be as it has a special meaning in query params. With this change, we'll make it so `+` is not encoded anymore.


VShingala (changes requested):
Let's add unit tests for the change before requesting review.

codegens/curl/lib/util.js
```diff
@@ -226,7 +226,8 @@ var self = module.exports = {
       .replace(/%5D/g, ']')
       .replace(/%7D/g, '}')
       .replace(/%25/g, '%')
-      .replace(/'/g, '%27');
+      .replace(/'/g, '%27')
+      .replace(/%2B/g, '+');
```
VShingala, line 230: @aman-v-singh Let's keep this at line no. 228. i.e. before we replace `%` chars.

aman-v-singh:
I included this change in the test and it is passing

Screenshot 2024-04-30 at 4 21 10 PM

VShingala (approved):
@aman-v-singh I think changes looks good. Let's make sure we test this behaviour in app as well before making a release.

VShingala (approved):

PR #745, opened 2024-05-24
Upgarde Restsharp to fix failing tests

aman-v-singh:
- Using Restsharp [version](https://restsharp.dev/docs/changelog#v111) `111.0.0` to fix failing tests.
- Client option `MaxTimeout` renamed to `Timeout` and changed type to `Timespan`


VShingala:
@aman-v-singh Not all users should have started using v11 of restsharp. Older users may still want the older behaviour of having `MaxTimeout`. In such cases, we should ideally introduce an option to configure the behaviour rather then enforcing new version to everyone. wdyt?

aman-v-singh:
@VShingala Makes sense. This issue happened because we did not specify the RestSharp version, so it used the latest version and did not support `MaxTimeout`. I have used the version (`v110.0.0`) that supports our code, and the tests are passing.

Issue #780, opened 2024-12-13
Regression in v1.11.0: detected package manager defaults to yarn


waldyrious:
The changes introduced in #744 and #765 result in the use of the incorrect package manager in some cases, where `yarn` is used where `npm` should be used instead.

For context, I am working in a repository that has a `package-lock.json` rather than a `yarn.lock` file; and indeed, my team uses `npm` instead of `yarn`. However, when we added `postman-code-generators` as a dependency, we started getting errors in CI, because we use an internal npm registry, and `yarn` is not able to access it using the same configurations and authentication data as `npm`.

For some reason, neither our top-level `package-lock.json`, nor this package's own [`package-lock.json`](https://github.com/postmanlabs/postman-code-generators/blob/develop/package-lock.json) are triggering the first condition that [detect-package-manager](https://github.com/egoist/detect-package-manager) was supposed to check (the presence of a given package manager's lockfile), so it goes to [step 2](https://github.com/egoist/detect-package-manager#how-does-this-work), which is to check what package managers are available in the machine. And indeed the CI runner we use has both `npm` and `yarn` available.

Unfortunately, in the presence of both `npm` and `yarn`, `detect-package-manager` gives precedence to `yarn`, for some reason. (This has been reported in https://github.com/egoist/detect-package-manager/issues/11, but after a year, there's no resolution yet — or even acknowledgement of the issue by the developer, for that matter).

@aman-v-singh, do you think it would be possible to reorder the array returned by `detect-package-manager`'s `detect()`, before it is used in [npm/deepinstall.js](https://github.com/postmanlabs/postman-code-generators/blob/develop/npm/deepinstall.js), so that `npm` comes first in the list?

waldyrious:
Thanks for responding! 

The project in question relies on Java dependencies, and therefore it uses Maven to orchestrate the build, including the step of running the `npm install` command (which, when it comes to the `postman-code-generators` dependency, runs the `npm/deepinstall.js` script which is what's swapping to `yarn` instead).

I'm not sure if this is relevant to why `detect-package-manager` isn't identifying the `package-lock.json` file,  but I thought I'd mention it just in case, because (1) I do know that Maven spawns commands in isolated shell environments (but not isolated working directories AFAIK), and (2) it explains why the log entries below are prefixed with `XX:XX:XX [INFO] [npm|docusaurus]`

Here's the relevant section of the CI job log:

```
[...]

PR #781, opened 2024-12-17
Removing lock files

aman-v-singh:
- Removed package lock and added files to not commit to .gitignore.
- This can help with fixing https://github.com/postmanlabs/postman-code-generators/issues/780


VShingala (approved):

VShingala:
@aman-v-singh There are other concerns also mentioned in [original issue](https://github.com/postmanlabs/postman-code-generators/issues/780#issuecomment-2548957811) that I'm not sure if we've addressed with this issue. It also makes sense to me. We can discuss more in original issue if there are any counter points to doing the change suggested there.

PR #799, opened 2025-10-25
Fix issues with deepinstall.js on windows

rtm516:
This fixes an issue with installing on windows due to `touch` not being available.

It also resolves a bug where the for loop can't return as expected causing next to be called 2x.
EG:
```
csharp-httpclient: yarn workspaces focus --all --production
Failed to run yarn install on codegen csharp-httpclient, here is the error:
csharp-restsharp: yarn workspaces focus --all --production
Failed to run yarn install on codegen csharp-restsharp, here is the error:
D:\...\node_modules\postman-code-generators\node_modules\async\dist\async.js:324
            if (fn === null) throw new Error("Callback was already called.");
                                   ^

Error: Callback was already called.
    at D:\...\node_modules\postman-code-generators\node_modules\async\dist\async.js:324:36
    at D:\...\node_modules\postman-code-generators\node_modules\async\dist\async.js:3646:17
    at D:\...\node_modules\postman-code-generators\npm\deepinstall.js:83:16
    at Array.forEach (<anonymous>)
    at D:\...\node_modules\postman-code-generators\npm\deepinstall.js:71:14
    at D:\...\node_modules\postman-code-generators\node_modules\async\dist\async.js:3641:28
    at replenish (D:\...\node_modules\postman-code-generators\node_modules\async\dist\async.js:446:21)
    at iterateeCallback (D:\...\node_modules\postman-code-generators\node_modules\async\dist\async.js:430:21)
    at D:\...\node_modules\postman-code-generators\node_modules\async\dist\async.js:327:20
    at D:\...\node_modules\postman-code-generators\node_modules\async\dist\async.js:3646:17
```


npm/deepinstall.js
```diff
@@ -44,7 +44,7 @@ async.series([
             command = 'yarn install --production --frozen-lockfile';
           }
           else {
-            command = 'touch yarn.lock && yarn workspaces focus --all --production'
```
shamasis, line 47: @VShingala  ... why would we even need to touch yarn.lock in the first place?

AI says: 
touch yarn.lock updates the file’s modification timestamp without changing its contents.
Old versions of Yarn (especially v2+ “Berry”) used mtime checks to decide whether to regenerate the .pnp.cjs file or re-link workspace dependencies.
If the yarn.lock looked older than .pnp.cjs or the cache, Yarn might skip re-processing it.
So, “touching” it ensures Yarn thinks “lockfile just changed — better re-resolve workspace deps.”

Don't think this is still relevant.

rtm516:
Added another fix

PR #802, opened 2025-11-12
chore: Update ubuntu version used in CI to v22 as ubuntu v20 is deprecated by GH Actions.

VShingala:
Ubuntu 20.04 runners were deprecated by GitHub Actions and are no longer available. As of 2024-2025, GitHub only supports:

ubuntu-22.04 (LTS)
ubuntu-24.04 (Latest LTS)

As 22.04 is current `ubuntu-latest`, we're moving to v22 wtih this change.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -11,6 +11,7 @@ jobs:
   Unit-Tests:
     runs-on: ubuntu-22.04
     strategy:
+      fail-fast: false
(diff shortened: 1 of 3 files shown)


shamasis (approved):

shamasis (approved):

shamasis (approved):
Tests failing for only 1 check. Because of httpbin not being available

PR #806, opened 2025-12-24
Fix postinstall script to support non-Yarn package managers

### LINKED ISSUE #792: Cant be used with non-yarn package managers and when yarn is also installed
**Describe the bug**
Cant be used in projects using other package managers other than yarn and when yarn is also installed 

**To Reproduce**
- (optional): npm install yarn -g
- mkdir bug && cd bug
- pnpm init
- pnpm add postman-code-generators 

work

**Screenshots**

Image

**Additional context**
https://github.com/postmanlabs/postman-code-generators/issues/792

sumomo015:
## Description

Fixes #792

This PR resolves an issue where the postinstall script fails when using package managers other than Yarn (such as pnpm or npm) in environments where Yarn is also installed globally.

## Problem

The current postinstall script in `npm/postinstall.js` attempts to detect the package manager being used, but fails to properly identify non-Yarn package managers when Yarn is installed globally. This causes the installation to fail with errors when users try to install `postman-code-generators` using pnpm or npm.

## Solution

Set the `INIT_CWD` environment variable during package manager detection to ensure the script correctly identifies the package manager being used, regardless of whether Yarn is installed globally.

## Reproduction Steps (Before Fix)

1. Create a Dockerfile with the following content: 

```dockerfile
FROM node:24.12.0

RUN npm install -g pnpm@10.26.2
RUN npm install -g yarn || true

WORKDIR /app

RUN pnpm init
# For demonstration only—enables 'dangerously-allow-all-builds', which is unsafe in production.
RUN echo "dangerously-allow-all-builds=true" >> .npmrc

# This step is expected to fail
RUN pnpm add postman-code-generators

CMD [ "echo", "Setup Complete" ]
```

2. Run the following command to build: 

```bash
docker build -t bug .
```

**Result:** The build fails when trying to install `postman-code-generators` with pnpm.

### DIFF
--- a/npm/deepinstall.js
+++ b/npm/deepinstall.js
@@ -9,7 +9,7 @@ var shell = require('shelljs'),
   fs = require('fs'),
   pwd = shell.pwd(),
   // eslint-disable-next-line no-process-env
-  initPwd = process.env.INIT_CWD;
+  initCwd = process.env.INIT_CWD;
 const args = process.argv,
   PATH_TO_CODEGENS_FOLDER = path.resolve(__dirname, '../codegens');
 
@@ -21,7 +21,7 @@ getSubfolders = (folder) => {
 
 async.series([
   function (next) {
-    detect({ cwd: initPwd }).then((res) => {
+    detect({ cwd: initCwd }).then((res) => {
       pm = res;
       console.log('Detected package manager: ' + pm);
       return next();
--- a/npm/deepinstall.js
+++ b/npm/deepinstall.js
@@ -7,7 +7,9 @@ var shell = require('shelljs'),
   command,
   getSubfolders,
   fs = require('fs'),
-  pwd = shell.pwd();
+  pwd = shell.pwd(),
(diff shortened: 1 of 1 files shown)


sumomo015:
As English is not my native language, this PR description was generated with the assistance of AI.

waldyrious:
I'd appreciate this! I had bumped into this issue before in #780 and #781 aimed to address it but I continue to face the issue (I mistakenly [thought](https://github.com/postmanlabs/postman-code-generators/issues/780#issuecomment-2553789153) it was fixed, but turns out it still happens occasionally, probably due to our cache system for CI images that doesn't always trigger the `deepinstall.js` script).

One alternative solution would be to swap [detect-package-manager](https://www.npmjs.com/package/detect-package-manager) for [package-manager-detector)](https://www.npmjs.com/package/package-manager-detector), since the former seems to be kind of abandoned and the latter has become significantly more popular despite being newer.

PR #810, opened 2026-03-09
Remove "path" dependency

opicacek:
This package has in the `package.json` specified:
```
  "engines": {
    "node": ">=18"
  }
```

📚  You should check docs about `lockfileVersion`: https://docs.npmjs.com/cli/v8/configuring-npm/package-lock-json#lockfileversion
Node v18 and above use NPM version v8+, which means they would always use `lockfileVersion: 3`.

----

Code changes in this PR:    
- Reformat package-lock file to `lockfileVersion: 3` format.
- Remove `path` as a 3rd party dependency. Use native `node:path` library instead. This custom `path` library is not compatible with Node v24+.
- Unlock versions of dependencies.
- Upgrade `eslint-plugin-mocha` to deal with incompatible peer dependencies.

### DIFF
--- a/codegens/php-guzzle/test/ci-install.sh
+++ b/codegens/php-guzzle/test/ci-install.sh
@@ -6,4 +6,4 @@ php -r "copy('https://getcomposer.org/installer', 'composer-setup.php');"
 php composer-setup.php
 php -r "unlink('composer-setup.php');"
 sudo mv composer.phar /usr/bin/composer
-composer global require guzzlehttp/guzzle:7.4.1
+composer global require guzzlehttp/guzzle:7.10.0
--- a/codegens/csharp-httpclient/npm-shrinkwrap.json
+++ b/codegens/csharp-httpclient/npm-shrinkwrap.json
@@ -1,5 +1,17 @@
 {
   "name": "@postman/codegen-csharp-httpclient",
   "version": "0.0.1",
-  "lockfileVersion": 1
+  "lockfileVersion": 3,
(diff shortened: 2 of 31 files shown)


codegens/php-guzzle/test/ci-install.sh
```diff
@@ -6,4 +6,4 @@ php -r "copy('https://getcomposer.org/installer', 'composer-setup.php');"
 php composer-setup.php
 php -r "unlink('composer-setup.php');"
 sudo mv composer.phar /usr/bin/composer
-composer global require guzzlehttp/guzzle:7.4.1
```
opicacek, line 9: Upgraded, because of:
```
Root composer.json requires guzzlehttp/guzzle 7.4.1 (exact version match: 7.4.1 or 7.4.1.0), found guzzlehttp/guzzle[7.4.1] but these were not loaded, because they are affected by security advisories ("PKSA-yfw5-9gnj-n2c7", "PKSA-k1b4-kshy-xgbh", "PKSA-2z36-j4q9-rsfy", "PKSA-fvw5-9t6n-nwvr", "PKSA-6d8m-6kgw-18zr"). Go to https://packagist.org/security-advisories/ to find advisory details.
```

VShingala: Makes sense! Thanks for this. Often time the codegen versions end up getting deprecated, causing CI issues over time.

VShingala (approved):
LGTM!

